PR #66, opened 2015-02-24
Handle recoverable exceptions in message processing workers.

wwajerowicz:


atlas-api/src/main/java/org/atlasapi/messaging/KafkaMessagingModule.java
```diff
@@ -37,7 +41,7 @@ public MessageSenderFactory messageSenderFactory() {
     @Override
     @Bean
     public MessageConsumerFactory<KafkaConsumer> messageConsumerFactory() {
-        return new KafkaMessageConsumerFactory(zookeeper, messagingSystem);
+        return new KafkaMessageConsumerFactory(zookeeper, messagingSystem, backOffIntervalMillis, maxBackOffMillis);
```
tfm: This should be a `Duration` not a `Long`

mbst-jamie:
All fine

atlas-processing/src/main/java/org/atlasapi/messaging/EquivalentScheduleStoreScheduleUpdateWorker.java
```diff
@@ -40,7 +41,7 @@ public void process(ScheduleUpdateMessage message) {
                 timer.stop();
             }
         } catch (WriteException e) {
-            log.error("update failed for " + message.getScheduleUpdate(), e);
+            throw new RecoverableException("update failed for " + message.getScheduleUpdate(), e);
```
tfm: Not sure it's up to the exception type to decide if it's recoverable or not. Shouldn't the exception type reflect the type of error, not whether it's recoverable or not?

tfm:
back to you, @wwajerowicz

atlas-processing/src/main/java/org/atlasapi/messaging/EquivalentScheduleStoreScheduleUpdateWorker.java
wwajerowicz: @tfm but doesn't the type of error actually reflect if it's recoverable or not?

PR #67, opened 2015-02-26
Send all stale broadcasts when updating schedules

wwajerowicz:


atlas-core/src/main/java/org/atlasapi/schedule/AbstractScheduleStore.java
```diff
@@ -36,6 +36,7 @@
 import com.metabroadcast.common.queue.MessagingException;
 import com.metabroadcast.common.time.DateTimeZones;
 import com.metabroadcast.common.time.Timestamp;
+import scala.concurrent.pilib;
```
tfm: ?

wwajerowicz: Sorry about that, that was imported by mistake, will remove

atlas-core/src/main/java/org/atlasapi/schedule/ScheduleBlockUpdater.java
```diff
@@ -23,7 +23,13 @@
 
 class ScheduleBlockUpdater {
 
-    public ScheduleBlocksUpdate updateBlocks(List<ChannelSchedule> currentBlocks, List<ItemAndBroadcast> updatedSchedule, Channel channel, Interval interval) {
+    public ScheduleBlocksUpdate updateBlocks(
+            List<ChannelSchedule> currentBlocks,
+            List<ChannelSchedule> pastBlocks,
```
tfm: It needs to be clear here that `pastBlocks` will be removed.

wwajerowicz: @tfm Does stale `staleBlocks` seem more reasonable?

PR #68, opened 2015-03-03
Add request data and license to the responses.

wwajerowicz:


atlas-api/src/main/java/org/atlasapi/LicenseModule.java
```diff
@@ -0,0 +1,25 @@
...
+@Configuration
+public class LicenseModule {
+
+    @Value("classpath:license.txt")
+    private org.springframework.core.io.Resource licenseText;
+
+    @Bean
+    EntityWriter<Object> licenseWriter() throws IOException {
+        return new LicenseWriter(
+                new License(IOUtils.toString(licenseText.getInputStream()))
```
mbst-jamie: May want to specify the CharSet, or use http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/io/CharStreams.html#toString(java.lang.Readable)

atlas-api/src/main/java/org/atlasapi/output/writers/ParamsMapWriter.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    @Nonnull
+    @Override
+    public String fieldName(Map<String, String[]> entity) {
+        return "params";
+    }
+    
+    private String getFirstOrEmptyString(String[] array) {
+        if(array.length > 0) {
```
mbst-jamie: return Iterables.getFirst(Arrays.asList(array), "");

mbst-jamie: or at least return (arr.length > 0 ? arr[0] : "");

PR #78, opened 2015-04-17
Use separate mongos in Processing for writing and reading.

wwajerowicz:


atlas-api/src/main/java/org/atlasapi/AtlasPersistenceModule.java
```diff
@@ -180,7 +189,7 @@ public IdGeneratorBuilder idGeneratorBuilder() {
 
             @Override
             public IdGenerator generator(String sequenceIdentifier) {
-                return new MongoSequentialIdGenerator(databasedMongo(), sequenceIdentifier);
+                return new MongoSequentialIdGenerator(databasedReadMongo(), sequenceIdentifier);
```
tfm: ID Generation - that writes to a database...

PR #80, opened 2015-04-21
Add count parameter to schedules

wwajerowicz:


atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleResolverBackedScheduleQueryExecutor.java
```diff
@@ -59,7 +60,14 @@ public ScheduleResolverBackedScheduleQueryExecutor(ChannelResolver channelResolv
             throws QueryExecutionException {
         
         List<Channel> channels = resolveChannels(query);
-        ListenableFuture<Schedule> schedule = scheduleResolver.resolve(channels, query.getInterval(), query.getSource());
+        if(!query.getEnd().isPresent()) {
+            throw new UnsupportedOperationException("ScheduleResolverBackedScheduleQueryExecutor doesn't support 'count' queries. Please specify 'to' parameter in your request");
```
mbst-jamie: ScheduleResolverBackedScheduleQueryExecutor means nothing to 99% of people who will read that error message

atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleRequestParser.java
```diff
@@ -170,14 +188,32 @@ private Id transformToId(String channelId) throws QueryParseException {
         }
     }
 
-    private Interval extractInterval(HttpServletRequest request) {
+    private DateTime extractFrom(HttpServletRequest request) {
         DateTime now = clock.now();
         DateTime from = dateTimeParser.parse(getParameter(request, "from"), now);
```
tfm: extract param names to constants?

atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleRequestParser.java
```diff
@@ -170,14 +188,32 @@ private Id transformToId(String channelId) throws QueryParseException {
...
+    private void checkInterval(DateTime from, DateTime to) {
         Interval queryInterval = new Interval(from, to);
         checkArgument(!queryInterval.toDuration().isLongerThan(maxQueryDuration), "Query interval cannot be longer than %s", maxQueryDuration);
-        return queryInterval;
+    }
+
+    private Integer extractCount(HttpServletRequest request) {
+        Integer count = Integer.valueOf(getParameter(request, "count"));
+
+        checkArgument(COUNT_RANGE.contains(count),  "'count' must be in range %s, was %s", COUNT_RANGE, count);
```
tfm: "in range" yet you're not providing a range; perhaps "'count' must be at most %s"? Also, shouldn't that be %d?

atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleResolverBackedScheduleQueryExecutor.java
tfm: ^ that

atlas-api/src/test/java/org/atlasapi/query/v4/schedule/ScheduleRequestParserTest.java
```diff
@@ -138,7 +143,56 @@ public void testCreatesSingleQueryFromValidSingleQueryStringWithNoExtension() th
...
         assertThat(query.getChannelId(), is(Id.valueOf(channel1.getId())));
-        assertThat(query.getInterval(), is(intvl));
+        assertThat(query.getStart(), is(start));
+        assertThat(query.getEnd().get(), is(end));
+        assertThat(query.getSource(), is(BBC));
+        assertThat(query.getContext().getAnnotations().forPath(ImmutableList.of(Resource.CONTENT)), is(Annotation.standard()));
+        assertThat(query.getContext().getApplicationSources(), is(sources));
+    }
+
+    @Test(expected=InvalidParameterException.class)
```
tfm: Can we also have a test for when count is too high? That would have caught the string format problem above.

tfm:
back to you @wwajerowicz

atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleRequestParser.java
wwajerowicz: @tfm it is in a way a range, we don't want people to specify negative numbers or 0 as count.

wwajerowicz: And also `checkArgument` only accepts `%s` in the template. Which works fine with integers.

tfm: The error needs to say "between 0 and x", then

wwajerowicz: does it really make sense to accept 0 for the count?

tfm: good point, 1 :)

atlas-api/src/main/java/org/atlasapi/query/common/SetBasedRequestParameterValidator.java
```diff
@@ -29,51 +30,35 @@ public Builder withOptionalParameters(String...parameters) {
             this.optional = ImmutableSet.copyOf(parameters);
             return this;
         }
+
+        public Builder withRequiredAlternativeParameters(String...parameters) {
```
tfm: Space after ...

atlas-api/src/test/java/org/atlasapi/query/v4/schedule/ScheduleRequestParserTest.java
wwajerowicz: Test added.

PR #82, opened 2015-04-27
Return all segment events for item.

wwajerowicz:


atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyContentTransformer.java
```diff
@@ -257,7 +257,10 @@ private SegmentEvent transformSegmentEvent(org.atlasapi.media.segment.SegmentEve
         se.setOffset(input.getOffset());
         se.setIsChapter(input.getIsChapter());
         org.atlasapi.media.entity.Description d = input.getDescription();
-        se.setDescription(new Description(d.getTitle(), d.getSynopsis(), d.getImage(), d.getThumbnail()));
```
tfm: Not sure about this; the `Description` object is a nicer approach. Perhaps just adding the interface `Sourced` is the right approach here?

PR #83, opened 2015-05-01
Merge victim should take into account requested ID

tfm:
If the requested ID has the same source as the most precedent
piece of content, then requested ID content should be the
merge victim, such that its attributes are taken in preference
to other equivalent content in the set.


wwajerowicz:
:+1:

PR #85, opened 2015-05-21
Player/Services port from Owl

mbst-jamie:


atlas-api/src/main/java/org/atlasapi/output/annotation/LocationsAnnotation.java
```diff
@@ -55,7 +66,14 @@ public void write(EncodedLocation entity, FieldWriter writer, OutputContext ctxt
...
             writer.writeList("available_countries", "country", policy.getAvailableCountries(), ctxt);
-            writer.writeField("service", policy.getService());
+            if (policy.getServiceRef() != null) {
+
+            }
+            if (policy.getPlayerRef() != null) {
+
+            }
+            writer.writeField("service", policy.getServiceRef());
+            writer.writeField("player", policy.getPlayerRef());
```
wwajerowicz: Shouldn't these two lines be moved inside the if statements?

mbst-jamie: Close, they shouldn't exist anymore, just hadn't finished working there yet ;p

atlas-api/src/main/java/org/atlasapi/output/annotation/LocationsAnnotation.java
```diff
@@ -55,7 +82,21 @@ public void write(EncodedLocation entity, FieldWriter writer, OutputContext ctxt
...
             writer.writeField("availability_end", policy.getAvailabilityEnd());
             writer.writeList("available_countries", "country", policy.getAvailableCountries(), ctxt);
-            writer.writeField("service", policy.getService());
+            if (policy.getServiceRef() != null) {
+                Optional<org.atlasapi.media.entity.Service> maybeService =
+                        serviceResolver.serviceFor(policy.getServiceRef().getId().longValue());
+                if (maybeService.isPresent()) {
+                    Service service = serviceTransformer.apply(maybeService.get());
+                    writer.writeObject(serviceWriter, service, ctxt);
+                }
```
wwajerowicz: If a Service which doesn't exist is reffered to, I think we should log it, not just ignore it. It's a data integrity issue,

atlas-api/src/main/java/org/atlasapi/output/annotation/LocationsAnnotation.java
```diff
@@ -55,7 +82,21 @@ public void write(EncodedLocation entity, FieldWriter writer, OutputContext ctxt
...
+            }
+            if (policy.getPlayerRef() != null) {
+                Optional<org.atlasapi.media.entity.Player> maybePlayer = playerResolver.playerFor(policy.getPlayerRef().getId().longValue());
+                if (maybePlayer.isPresent()) {
+                    Player player = playerTransformer.apply(maybePlayer.get());
+                    writer.writeObject(playerWriter, player, ctxt);
```
wwajerowicz: Same with Player

wwajerowicz:
Other than two comments on the code, :+1:

PR #92, opened 2015-06-01
Remove no longer published content from the idx

mbst-jamie:


atlas-elasticsearch/src/main/java/org/atlasapi/content/EsContentIndex.java
```diff
@@ -430,6 +439,38 @@ protected Void visitContainer(Container container) {
         }
     }
 
+    private void unindexNoLongerPublishedContent(Content content) {
```
tfm: perhaps just `removeFromIndex(Content content)`

tfm:
Test?

tfm:
Back to you, @mbst-jamie

atlas-elasticsearch/src/main/java/org/atlasapi/content/EsContentIndex.java
mbst-jamie: This can also be de-duped a bit... ty

PR #101, opened 2015-06-16
Port fuzzy search (q param) to content endpoint (poorly)

mbst-jamie:


atlas-api/src/main/java/org/atlasapi/query/common/QueryContextParser.java
```diff
@@ -57,7 +57,7 @@ public QueryContext parseListContext(HttpServletRequest request) throws QueryPar
             userFetcher.getParameterNames(),
             annotationExtractor.getParameterNames(), 
             selectionBuilder.getParameterNames(),
-            ImmutableSet.of(JsonResponseWriter.CALLBACK, "order_by"))));
+            ImmutableSet.of(JsonResponseWriter.CALLBACK, "order_by", "q", "title_boost"))));
```
mbst-jamie: title_boost still doesnt work on the API due to being interpreted as a title attribute with an operator of boost...

PR #102, opened 2015-06-17
Mbst 11643 ingest priority add to model and search index

mbst-jamie:


atlas-api/src/main/java/org/atlasapi/query/common/IndexQueryParser.java
```diff
@@ -0,0 +1,78 @@
...
+
+import org.atlasapi.content.IndexQueryParams;
+import org.atlasapi.content.QueryOrdering;
+import org.atlasapi.content.FuzzyQueryParams;
+
+import java.util.Map;
+import java.util.Optional;
+
+
+/***
```
tsprlng: is this a TODO???

atlas-api/src/main/java/org/atlasapi/query/common/IndexQueryParser.java
```diff
@@ -0,0 +1,78 @@
...
+        );
+    }
+
+    private Optional<QueryOrdering> orderingFrom(Query<?> query) throws QueryParseException {
+        Map params = query.getContext().getRequest().getParameterMap();
+        if (!params.containsKey("order_by")) {
+            return Optional.empty();
+        }
+        String[] orderByVals = ((String[]) params.get("order_by"));
+        if (orderByVals[0] == null) {
```
tsprlng: Iterables.getOnlyOne

tsprlng: This will surely just fail on [0] if it's empty? Or is it actually going to be a one-element list of [null]? What a crazy world.

mbst-jamie: Itll IOOBE, you're right

mbst-jamie: Cant getOnlyElement on an array tho

tsprlng:
:+1:

PR #107, opened 2015-06-17
Region customisation

wwajerowicz:


atlas-api/src/main/java/org/atlasapi/output/OutputContext.java
```diff
@@ -24,26 +26,45 @@
...
     
     private final ActiveAnnotations annotations;
     private final ApplicationSources applicationSources;
     private final List<Resource> resources;
     private final HttpServletRequest request;
+    private final Optional<Region> region;
 
     public OutputContext(ActiveAnnotations activeAnnotations,
-                         ApplicationSources applicationSources, HttpServletRequest request) {
+                         ApplicationSources applicationSources, HttpServletRequest request, Region region) {
```
mbst-jamie: Have it pass in an Optional<Region> rather than accept null, or atleast mark it as @Nullable

mbst-jamie:
:+1:

PR #117, opened 2015-07-02
hasTitleArt and Source on Images

mbst-jamie:


tfm:
Can haz test added in `OutputContentMerger`?

PR #130, opened 2015-08-03
Order based on sort key for sub items

mbst-jamie:


tfm:
Can we have a test please?

mbst-jamie:
test added to master

PR #132, opened 2015-08-03
Paginate over subitems through query params

mbst-jamie:


atlas-api/src/main/java/org/atlasapi/output/annotation/SubItemAnnotation.java
```diff
@@ -25,12 +27,23 @@ public SubItemAnnotation(NumberToShortStringCodec idCodec) {
     @Override
     public void write(Content entity, FieldWriter writer, OutputContext ctxt) throws IOException {
         if (entity instanceof Container) {
+
+            Integer offset = 0;
+            Integer limit = Integer.MAX_VALUE;
+            if (!Strings.isNullOrEmpty(ctxt.getRequest().getParameter("sub_item.offset"))) {
```
tfm: Pull into a variable first - otherwise the duplication will hurt when you check one thing in the conditional, and then do `Integer.valueOf` of another. Also pull `sub_item.offset` into a constant. 

Shouldn't it be `sub_items` to be consistent with the key?

mbst-jamie: Bug

atlas-api/src/main/java/org/atlasapi/output/annotation/SubItemAnnotation.java
```diff
@@ -25,12 +29,25 @@ public SubItemAnnotation(NumberToShortStringCodec idCodec) {
     @Override
     public void write(Content entity, FieldWriter writer, OutputContext ctxt) throws IOException {
         if (entity instanceof Container) {
+
+            Integer offset = 0;
+            Integer limit = Integer.MAX_VALUE;
+            String offsetStr = ctxt.getRequest().getParameter(SUB_ITEM_OFFSET);
+            if (!com.google.common.base.Strings.isNullOrEmpty(offsetStr)) {
```
tfm: Why the fq use?

atlas-api/src/main/java/org/atlasapi/output/annotation/SubItemAnnotation.java
```diff
@@ -25,12 +29,25 @@ public SubItemAnnotation(NumberToShortStringCodec idCodec) {
...
+                limit = Integer.valueOf(limitStr);
+            }
+
             Container container = (Container) entity;
             ImmutableList<ItemRef> orderedRefs = Ordering.natural()
                     .onResultOf(ItemRef::getSortKey)
                     .reverse()
                     .nullsLast()
                     .immutableSortedCopy(container.getItemRefs());
+            orderedRefs = orderedRefs.stream().skip(offset).limit(limit).collect(ImmutableCollectors.toList());
```
tfm: multi-line would be easier to read?

atlas-api/src/main/java/org/atlasapi/output/annotation/SubItemSummariesAnnotations.java
```diff
@@ -24,12 +28,24 @@ public SubItemSummariesAnnotations(SubItemSummaryListWriter subItemSummaryListWr
...
+                limit = Integer.valueOf(limitStr);
+            }
+
             Container container = (Container) entity;
             ImmutableList<ItemSummary> summaries = Ordering.natural()
                     .onResultOf((ItemSummary summary) -> summary.getItemRef().getSortKey())
                     .reverse()
                     .nullsLast()
                     .immutableSortedCopy(container.getItemSummaries());
+            summaries = summaries.stream().skip(offset).limit(limit).collect(ImmutableCollectors.toList());
```
tfm: multi-line easier to read

atlas-api/src/main/java/org/atlasapi/output/annotation/SubItemAnnotation.java
mbst-jamie: you're asking all the hard questions

PR #133, opened 2015-08-05
Separate precedence for hierarchies

tfm:
Allow precedence ordering for merging of containers'
contents to be specified separately from the overall precedence
ordering.

### DIFF
--- a/atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
+++ b/atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
@@ -10,6 +10,7 @@ import javax.annotation.Nullable;
 import com.google.common.collect.ImmutableMap;
 
 import org.atlasapi.application.ApplicationSources;
+import org.atlasapi.content.Brand;
 import org.atlasapi.content.Broadcast;
 import org.atlasapi.content.Certificate;
 import org.atlasapi.content.Clip;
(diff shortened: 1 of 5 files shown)


atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -485,20 +487,30 @@ public void mergeIn(ApplicationSources sources, Container chosen, Iterable<Conta
...
+        Optional<Ordering<Sourced>> sourcedContentHierarchyOrdering = sources.getSourcedContentHierarchyOrdering();
+        if (sourcedContentHierarchyOrdering.isPresent()) {
+            orderedEquivalents = sourcedContentHierarchyOrdering.get().sortedCopy(Iterables.concat(ImmutableSet.of(chosen), notChosen));
+        } else {
+            orderedEquivalents = Iterables.concat(ImmutableSet.of(chosen), notChosen);
+        }
+        
+        Iterable<Container> contentHierarchySourceOrderedContainers = Iterables.filter(orderedEquivalents, Container.class);
         if (chosen.getUpcomingContent().isEmpty()) {
             chosen.setUpcomingContent(
```
wwajerowicz: `chosen` is selected using `sourceReadPrecedence`. If `chosen` has upcoming content, it's upcoming content will be selected, which will be incosistent with `sourcedContentHierarchyOrdering`. We either need to use  source read precedence ordering or not use `chosen`.

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -485,20 +487,30 @@ public void mergeIn(ApplicationSources sources, Container chosen, Iterable<Conta
...
-                            notChosenContainers,
+                            contentHierarchySourceOrderedContainers,
                             input -> input.getUpcomingContent().isEmpty() ? null : input.getUpcomingContent(),
                             ImmutableMap.of()
                     )
             );
         }
+        
         if(chosen.getItemSummaries().isEmpty()) {
             chosen.setItemSummaries(
```
wwajerowicz: Same here.

wwajerowicz:
also `ApplicationModelTranslatorTest` is broken

PR #141, opened 2015-09-07
Equivalent es index

mbst-jamie:


atlas-elasticsearch/src/main/java/org/atlasapi/content/PseudoEquivalentEsContentIndex.java
```diff
@@ -0,0 +1,70 @@
...
+
+            IndexQueryResult result = Futures.get(
+                    delegate.query(query, publishers, selectionForDelegate, searchParam),
+                    Exception.class
+            );
+
+            FluentIterable<Id> ids = result.getIds();
+
+            ImmutableMap<Long, Long> equivIndexResults =
+                    Futures.get(equivalenceIndex.lookup(ids.transform(Id::longValue)), Exception.class);
```
mbst-jamie: Does this want a cache? Does it want some clever delegate for preventing the resolution of 10k+ IDs at once?

wwajerowicz:
Couldn't we store the equivalent set id together with the content in ES, rather than resolve it every time?

PR #146, opened 2015-09-22
Resolve from Owl if Content not present in Deer in while updating Equ…

wwajerowicz:
…ivalent content.

### DIFF
--- a/atlas-processing/src/main/java/org/atlasapi/messaging/EquivalentContentStoreContentUpdateWorker.java
+++ b/atlas-processing/src/main/java/org/atlasapi/messaging/EquivalentContentStoreContentUpdateWorker.java
@@ -4,8 +4,14 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import javax.annotation.Nullable;
 
+import com.google.common.collect.ImmutableList;
+import com.google.common.util.concurrent.Futures;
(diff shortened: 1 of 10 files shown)


atlas-processing/src/main/java/org/atlasapi/messaging/EquivalentContentStoreContentUpdateWorker.java
```diff
@@ -18,27 +24,39 @@
...
     public void process(ResourceUpdatedMessage message) throws RecoverableException {
         try {
-            if (messageTimer != null) {
-                Timer.Context timer = messageTimer.time();
-                equivalentContentStore.updateContent(message.getUpdatedResource());
-                timer.stop();
-            } else {
-                equivalentContentStore.updateContent(message.getUpdatedResource());
-            }
+            Timer.Context timer = messageTimer.time();
```
tfm: Why is this now null-safe?

tfm:
One comment, otherwise :+1:

atlas-processing/src/main/java/org/atlasapi/messaging/EquivalentContentStoreContentUpdateWorker.java
wwajerowicz: We're always injecting it so not sure why there were these null checks before

PR #151, opened 2015-09-30
Implemented indexing listener and worker to index content updated by …

Dalamar42:
…the equivalent content store

- Added ContentRef field to EquivalentContentUpdatedMessage.java and updated Jackson serialization
- Created EquivalentContentIndexingWorker.java and added listener using it to WorkersModule.java that listens to the equivalent content changes queue


tfm:
:+1:

PR #164, opened 2015-10-22
Delete stale broadcasts only from update interval.

wwajerowicz:


atlas-cassandra/src/main/java/org/atlasapi/schedule/CassandraEquivalentScheduleStore.java
```diff
@@ -451,10 +452,10 @@ private ByteBuffer serialize(Broadcast broadcast) {
         return ByteBuffer.wrap(broadcastSerializer.serialize(broadcast).build().toByteArray());
     }
 
-    private List<RegularStatement> deleteStatements(Publisher src, Set<BroadcastRef> staleBroadcasts) {
+    private List<RegularStatement> deleteStatements(Publisher src, Set<BroadcastRef> staleBroadcasts, Interval interval) {
```
tfm: Think we need some documentation here on which interval, and why it matters.

atlas-testlib/src/main/java/org/atlasapi/schedule/EquivalentScheduleStoreTester.java
```diff
@@ -1053,7 +1053,125 @@ public void testUpdatingAScheduleRemovesStaleBroadcastsEvenIfNotPresentInStaleBr
         );
 
     }
-    
+
+
+    public void testDoesnRemoveStaleBroadcastsFromOutsideInterval() throws Exception {
```
tfm: Can you add some documentation which explains this test case (probably pretty much the comment in chat at 0927 this morning)

tfm:
Back to you, @wwajerowicz

PR #169, opened 2015-10-26
Changed bootstrap/debug controllers to not send messages to Kafka

Dalamar42:
- Created null message sending equivalent content store and wired into ContentBootstrapController.java and ContentDebugController.java


JamieAP:
LGTM

PR #170, opened 2015-10-27
Feature/POST content endpoint

emilssolmanis:
Adds basic endpoint to ingest `Content` via Deer.


atlas-api/src/main/java/org/atlasapi/util/ClientModelConverter.java
```diff
@@ -0,0 +1,109 @@
...
+            item.setId(codec.decode(api.getId()).longValueExact());
+        }
+
+        // TODO: immutable collector
+        if (api.getAliases() != null) {
+            item.setAliases(api
+                            .getAliases()
+                            .stream()
+                            .map(clientAlias -> new Alias(clientAlias.getNamespace(), clientAlias.getValue()))
+                            .collect(Collectors.toList())
```
Dalamar42: atlas-deer should already have ImmutableCollectors somewhere I think

atlas-api/src/test/java/org/atlasapi/util/ClientModelConverterTest.java
```diff
@@ -0,0 +1,207 @@
...
+    @Before
+    public void setUp() {
+        converter = new ClientModelConverter();
+    }
+
+    @Test
+    public void convertsContent() {
+        /*
+          Still inconsistent:
+            DisplayTitle displayTitle, -- doesn't really exist on Content, is calculated
```
Dalamar42: If that field doesn't exist in Content then I think it should be deleted from the client model.

atlas-api/src/test/java/org/atlasapi/util/ClientModelConverterTest.java
```diff
@@ -0,0 +1,207 @@
...
+            String priority, -- it's a compound object, not a String
```
Dalamar42: That has to do with changes Martins is making. Check with him

atlas-api/src/test/java/org/atlasapi/util/ClientModelConverterTest.java
```diff
@@ -0,0 +1,207 @@
...
+            List<String> certificates, -- has a country too
```
Dalamar42: Same for the rest. If you are seeing differences then I think the model should be fixed.

Dalamar42:
Just a couple of comments. Other than that looks good to me.

atlas-api/src/test/java/org/atlasapi/util/ClientModelConverterTest.java
emilssolmanis: @Dalamar42 tl; dr is Tom said to not bother right now as we only need a basic `Content` ingestion thing and should fix as we stumble upon things. I'm generally of the mindset of "do once and forget about it later" though, so I guess we can have a chat about it.

Dalamar42: @emilssolmanis I tend to be of the same mindset. Let's have a couple min chat tomorrow.

PR #176, opened 2015-11-03
Parallelise requests to C* rather than use batching

Dalamar42:


JamieAP:
LGTM

Dalamar42:
This can not be merged as we rely on the batch operations to ensure consistency of the equivalent_content table

PR #177, opened 2015-11-03
Remove unnecessary graph column from equivalent_content store

Dalamar42:
- Graph column is only being used to ensure consistency between the
equivalent_content and equivalent_content_index stores. However we
accept eventual consistency anyway and also if the data is
inconsistent the existing code does not log or attempt to resolve the
issue. Therefore it is safe to stop using this column.
- Large batch updates to C* that are due to this graph column are a
likely cause of C* instability
- Remove duplicate updates of equivalent_content rows


JamieAP:
LGTM

PR #178, opened 2015-11-10
Mbst 13222 adding separate score reasons

MartinsIrbe:
Added changes to common.proto and content.proto to have two separate list for Priority score reasons.
Furthermore updated following classes for serialisation and deserialisation:
PrioritySerializer
ContentSerializationVisitor
ContentDeserializationVisitor
Updated tests:
PrioritySerializer

### DIFF
--- a/atlas-cassandra/src/main/java/org/atlasapi/content/ContentSerializationVisitor.java
+++ b/atlas-cassandra/src/main/java/org/atlasapi/content/ContentSerializationVisitor.java
@@ -133,18 +133,19 @@ public final class ContentSerializationVisitor implements ContentVisitor<Builder
         for (RelatedLink relatedLink : content.getRelatedLinks()) {
             builder.addRelatedLink(relatedLinkSerializer.serialize(relatedLink));
         }
-        if (content.getPriority() != null) {
+        Priority priority = content.getPriority();
+        if (priority != null) {
             ContentProtos.Priority.Builder priorityBuilder = ContentProtos.Priority.newBuilder();
-            Priority priority = content.getPriority();
             if (priority.getPriority() != null) {
                 priorityBuilder.setScore(priority.getPriority());
-                if (priority.getReasons() != null) {
-                    if (priority.getReasons().getPositive() != null) {
-                        priorityBuilder.addAllPositiveReasons(priority.getReasons().getPositive());
-                    }
-                    if (priority.getReasons().getNegative() != null) {
-                        priorityBuilder.addAllNegativeReasons(priority.getReasons().getNegative());
-                    }
+            }
+            PriorityScoreReasons scoreReasons = priority.getReasons();
+            if (scoreReasons != null) {
+                if (scoreReasons.getPositive() != null) {
+                    priorityBuilder.addAllPositiveReasons(scoreReasons.getPositive());
+                }
+                if (scoreReasons.getNegative() != null) {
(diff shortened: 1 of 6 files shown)


atlas-cassandra/src/main/java/org/atlasapi/content/ContentSerializationVisitor.java
```diff
@@ -137,10 +137,15 @@ private Builder visitDescribed(Described content) {
             ContentProtos.Priority.Builder priorityBuilder = ContentProtos.Priority.newBuilder();
             Priority priority = content.getPriority();
             if (priority.getPriority() != null) {
-                priorityBuilder.addAllReasons(priority.getReasons());
-            }
-            if (priority.getPriority() != null) {
                 priorityBuilder.setScore(priority.getPriority());
+                if (priority.getReasons() != null) {
```
emilssolmanis: Extract `priority.getReasons()` to a var, you're using it at least 6 times here.

atlas-cassandra/src/main/java/org/atlasapi/content/PrioritySerializer.java
```diff
@@ -10,8 +10,15 @@
         CommonProtos.Priority.Builder builder = CommonProtos.Priority.newBuilder();
 
         if(source.getReasons() != null) {
-            for (String reason : source.getReasons()) {
-                builder.addReason(reason);
+            if (source.getReasons().getPositive() != null) {
```
emilssolmanis: Same as above, extract as var.

atlas-cassandra/src/main/protobuf/common.proto
```diff
@@ -118,8 +118,9 @@ message Synopses {
 }
 
 message Priority {
-    repeated string reason = 1;
+    repeated string positive_reasons = 1;
```
emilssolmanis: Reusing field IDs is a Very Bad Idea™ unless the fields are actually the same. Are you sure this is what you want? Doesn't seem like `reason` is the same field semantically as `positive_reasons`. For that matter, what happens to compatibility in this case?

atlas-cassandra/src/main/protobuf/content.proto
```diff
@@ -119,7 +119,8 @@ message Restriction {
 
 message Priority {
     optional double score = 1;
-    repeated string reasons = 2;
+    repeated string positive_reasons = 2;
```
emilssolmanis: Same as above.

atlas-cassandra/src/main/protobuf/common.proto
MartinsIrbe: I will update and mark `reason` as deprecated.

atlas-cassandra/src/main/protobuf/content.proto
MartinsIrbe: :+1:

atlas-cassandra/src/main/java/org/atlasapi/content/ContentSerializationVisitor.java
MartinsIrbe: :+1:

atlas-cassandra/src/main/java/org/atlasapi/content/PrioritySerializer.java
MartinsIrbe: :+1:

emilssolmanis:
:+1: :shipit:

PR #179, opened 2015-11-10
added advertiseFrom field to the channel and annotatioons

Spartakovic:


Dalamar42:
You need to also update:
- the Deer version of the model
- the code the transforms the Owl model to the Deer model
- the protocol buffers serialisation and deserialisation to ensure this is stored in C*

Dalamar42:
LGTM

atlas-api/src/main/java/org/atlasapi/output/annotation/ChannelAdvertisedAnnotation.java
```diff
@@ -0,0 +1,32 @@
...
+
+public class ChannelAdvertisedAnnotation extends OutputAnnotation<Channel> {
+    private final ChannelWriter channelWriter;
+
+    public ChannelAdvertisedAnnotation(ChannelWriter channelWriter) {
+        this.channelWriter = checkNotNull(channelWriter);
+    }
+
+    @Override
+    public void write(Channel entity, FieldWriter format, OutputContext ctxt) throws IOException {
```
tfm: These needs to honour the existing logic for channel output, too, which is based on start date of the channel numbering. Does that happen elsewhere already?

atlas-api/src/main/java/org/atlasapi/output/annotation/ChannelAdvertisedAnnotation.java
```diff
@@ -0,0 +1,32 @@
...
+        DateTime advertiseFrom = entity.getAdvertiseFrom();
+        if (advertiseFrom != null && isInFuture(advertiseFrom)) {
```
tfm: This is incorrect, should be

`if (advertiseFrom == null || !isInFuture(advertiseFrom))`

tfm:
No tests

Dalamar42:
Update the LegacyChannelTransformer test to check you are correctly transforming the new field.

Dalamar42:
See if ChannelWriter has a test class. If it does add a test there. Otherwise LGTM

PR #182, opened 2015-11-12
MBST-13263: Added priority score reasons for the output json object.

MartinsIrbe:


atlas-api/src/main/java/org/atlasapi/output/annotation/ExtendedDescriptionAnnotation.java
```diff
@@ -20,11 +20,7 @@
 import org.atlasapi.entity.ResourceRef;
 import org.atlasapi.output.FieldWriter;
 import org.atlasapi.output.OutputContext;
-import org.atlasapi.output.writers.CertificateWriter;
-import org.atlasapi.output.writers.LanguageWriter;
-import org.atlasapi.output.writers.ReleaseDateWriter;
-import org.atlasapi.output.writers.RestrictionWriter;
-import org.atlasapi.output.writers.SubtitleWriter;
+import org.atlasapi.output.writers.*;
```
Spartakovic: not good

atlas-api/src/main/java/org/atlasapi/output/annotation/ExtendedDescriptionAnnotation.java
```diff
@@ -44,6 +41,7 @@ public ExtendedDescriptionAnnotation() {
         this.subtitleWriter = new SubtitleWriter(languageWriter);
         releaseDateWriter = new ReleaseDateWriter();
```
Spartakovic: why it not this.releaseDateWriter?

atlas-api/src/main/java/org/atlasapi/output/annotation/ExtendedDescriptionAnnotation.java
MartinsIrbe: updated

atlas-api/src/main/java/org/atlasapi/output/annotation/ExtendedDescriptionAnnotation.java
MartinsIrbe: Not sure as I haven't added this.

atlas-api/src/main/java/org/atlasapi/output/annotation/ExtendedDescriptionAnnotation.java
timothy22000: Martins, check your IntelliJ settings to not let it import * when there is more than 5 imports from the same place. Ask me tomorrow if you haven't found out about it.

PR #185, opened 2015-11-13
Allow for bootstrapping source plus equivalents to that source

Dalamar42:
- ContentBootstrapListener can optionally bootstrap equivalents
- Create null message sending equivalence graph store


tfm:
:+1:

PR #186, opened 2015-11-16
Added endpoint for deactivating third party credentials

Spartakovic:
We need an endpoint that we can `POST` to which will deactivate a users third-party (Google/Github/Twitter) account from atlas admin.

### DIFF
--- a/atlas-api/src/main/java/org/atlasapi/application/users/UsersController.java
+++ b/atlas-api/src/main/java/org/atlasapi/application/users/UsersController.java
@@ -63,7 +63,7 @@ public class UsersController {
             NumberToShortStringCodec idCodec,
             UserFetcher userFetcher,
             UserStore userStore,
-            MongoDBCredentialsStore credentialsStore,
+            CredentialsStore credentialsStore,
             Clock clock) {
         this.requestParser = requestParser;
         this.queryExecutor = queryExecutor;
@@ -73,7 +73,7 @@ public class UsersController {
         this.userFetcher = userFetcher;
         this.userStore = userStore;
         this.clock = clock;
-        this.credentialsStore = credentialsStore;
+        this.credentialsStore = (MongoDBCredentialsStore) credentialsStore;
     }
 
     @RequestMapping({ "/4/users/{uid}.*", "/4/users.*" })
@@ -163,6 +163,7 @@ public class UsersController {
         ResponseWriter writer = null;
         try {
             User editingUser = userFetcher.userFor(request).get();
+
             if (!editingUser.is(Role.ADMIN)) {
                 throw new ResourceForbiddenException();
             }
--- a/atlas-api/src/main/java/org/atlasapi/application/users/UsersController.java
+++ b/atlas-api/src/main/java/org/atlasapi/application/users/UsersController.java
@@ -63,7 +63,7 @@ public class UsersController {
             NumberToShortStringCodec idCodec,
             UserFetcher userFetcher,
             UserStore userStore,
-            CredentialsStore credentialsStore,
+            MongoDBCredentialsStore credentialsStore,
             Clock clock) {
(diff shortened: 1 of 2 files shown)


atlas-api/src/main/java/org/atlasapi/application/users/UsersController.java
```diff
@@ -146,6 +155,33 @@ public void updateUser(HttpServletRequest request,
             new ErrorResultWriter().write(summary, writer, request, response);
         }
     }
+
+    @RequestMapping(value = "/4/users/deactivate/{uid}.*", method = RequestMethod.POST)
+    public void deactivateUser(HttpServletRequest request,
+            HttpServletResponse response,
+            @PathVariable String uid) throws IOException {
+        ResponseWriter writer = null;
```
JamieAP: This is unused, so why make it?

JamieAP:
Otherwise LGTM

atlas-api/src/main/java/org/atlasapi/application/users/UsersController.java
Spartakovic: it is used for ErrorSummary)

atlas-api/src/main/java/org/atlasapi/application/users/UsersController.java
```diff
@@ -146,6 +155,32 @@ public void updateUser(HttpServletRequest request,
...
+        try {
+            User editingUser = userFetcher.userFor(request).get();
+            if (!editingUser.is(Role.ADMIN)) {
+                throw new ResourceForbiddenException();
+            }
+            Id userId = Id.valueOf(idCodec.decode(uid));
```
cognitive-i: Some whitespace would be helpful here.

atlas-api/src/main/java/org/atlasapi/application/users/UsersController.java
```diff
@@ -146,6 +155,32 @@ public void updateUser(HttpServletRequest request,
...
+    @RequestMapping(value = "/4/users/deactivate/{uid}.*", method = RequestMethod.POST)
+    public void deactivateUser(HttpServletRequest request,
+            HttpServletResponse response,
+            @PathVariable String uid) throws IOException {
+        ResponseWriter writer = null;
+        try {
+            User editingUser = userFetcher.userFor(request).get();
+            if (!editingUser.is(Role.ADMIN)) {
+                throw new ResourceForbiddenException();
+            }
```
cognitive-i: Some whitespace would be helpful here

atlas-api/src/main/java/org/atlasapi/application/users/UsersController.java
```diff
@@ -65,6 +73,7 @@ public UsersController(UserAwareQueryParser<User> requestParser,
         this.userFetcher = userFetcher;
         this.userStore = userStore;
         this.clock = clock;
+        this.credentialsStore = (MongoDBCredentialsStore) credentialsStore;
```
emilssolmanis: Why do we need the cast? Should we actually be fixing the interface?

atlas-api/src/main/java/org/atlasapi/application/users/UsersController.java
```diff
@@ -146,6 +155,32 @@ public void updateUser(HttpServletRequest request,
...
+                Credentials credentials = credentialsStore.find(userRef).requireValue();
+                credentialsStore.expired(credentials.authToken());
+            } else {
+                throw new NotFoundException(userId);
+            }
+        } catch (Exception e) {
+            log.error("Request exception " + request.getRequestURI(), e);
+            ErrorSummary summary = ErrorSummary.forException(e);
+            new ErrorResultWriter().write(summary, writer, request, response);
+            }
```
emilssolmanis: Misaligned whitespace.

PR #187, opened 2015-11-16
Only delete stale ES entries of a different type than the current one

Dalamar42:
- When deleting stale ES entries do not delete entries of the
same type as the currently indexed Content to avoid losing data
denormalised on series by episodes, e.g. locations and broadcasts


JamieAP:
:+1:

PR #188, opened 2015-11-17
Change ES broadcast range filter to cover broadcasts partially in range

Dalamar42:
- Current actionable filter will fail to return currently broadcast
content if the broadcastTimeGreaterThan time is the current time.
This will ensure that if a broadcast is even partially in the
requested range it will show up in the query
- Added defensive guard to ensure broadcastTimeGreaterThan is less or
 equal to broadcastTimeLessThan if both have been provided


JamieAP:
LGTM

PR #189, opened 2015-11-17
Switch to using multiple getKey's instead of the getKeySlice

emilssolmanis:


Dalamar42:
LGTM

Dalamar42:
Agree on timeout

Dalamar42:
Also it doesn't apply here but for future things, there is something in Futures to convert a collection of listenables returning single values to a listenable returning a collection of the same return type

emilssolmanis:
@Dalamar42 err, yes, it's `Futures.allAsList` and it's used on line 169.

PR #190, opened 2015-11-17
Use prepared statements in Datastax C*

emilssolmanis:


emilssolmanis:
**NOTE**: there's also a variant of `prepareAsync` which returns a `ListenableFuture`, as it stands the impl does the `prepare` calls sync. I don't think that's an issue since it's a one-off thing, but if you think it is, feel free to flag it.

atlas-cassandra/src/main/java/org/atlasapi/content/CassandraEquivalentContentStore.java
```diff
@@ -273,46 +292,44 @@ private void deleteStaleRows(EquivalenceGraph updated, ImmutableSet<EquivalenceG
...
-        session.execute(batch(deletes.toArray(new RegularStatement[deletes.size()]))
-                .setConsistencyLevel(writeConsistency));
+
+        BatchStatement batchStatement = new BatchStatement();
+        batchStatement.setConsistencyLevel(writeConsistency);
+
+        batchStatement.addAll(created.stream()
+                .flatMap(graph -> graph.getEquivalenceSet().stream())
+                .map(elem -> rowDelete.bind()
+                        .setLong("setId", id)
```
tfm: moved from constants to in-line strings?

tfm:
only one minor thing, otherwise :+1:

emilssolmanis:
@tfm not really, the strings are arg placeholder names, can make them all constants if you feel like that'll be better.

PR #192, opened 2015-11-18
Mbst 13131 configure cassandra connection timeouts

cognitive-i:
Extracted builder from DataStax to permit configuration of timeouts via environment.properties

### DIFF
--- a/atlas-api/src/main/java/org/atlasapi/AtlasPersistenceModule.java
+++ b/atlas-api/src/main/java/org/atlasapi/AtlasPersistenceModule.java
@@ -133,10 +133,10 @@ public class AtlasPersistenceModule {
         // this code is borrowed from DatastaxCassandraService::DatastaxCassandraService
         // but enhances timeout criteria to improve Atlas performance when cluster is loaded
         PoolingOptions poolingOptions = new PoolingOptions();
-        poolingOptions.setMaxConnectionsPerHost(HostDistance.LOCAL, ((Integer) Preconditions.checkNotNull(connectionsPerHostLocal)).intValue());
-        poolingOptions.setCoreConnectionsPerHost(HostDistance.LOCAL, ((Integer) Preconditions.checkNotNull(connectionsPerHostLocal)).intValue());
-        poolingOptions.setMaxConnectionsPerHost(HostDistance.REMOTE, ((Integer) Preconditions.checkNotNull(connectionsPerHostRemote)).intValue());
-        poolingOptions.setCoreConnectionsPerHost(HostDistance.REMOTE, ((Integer) Preconditions.checkNotNull(connectionsPerHostRemote)).intValue());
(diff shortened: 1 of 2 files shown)


atlas-api/src/main/java/org/atlasapi/AtlasPersistenceModule.java
```diff
@@ -6,6 +6,10 @@
 
 import javax.annotation.PostConstruct;
 
+import com.datastax.driver.core.*;
```
emilssolmanis: Misonfigged IntelliJ? We don't normally do star imports.

atlas-api/src/main/java/org/atlasapi/AtlasPersistenceModule.java
```diff
@@ -120,6 +126,35 @@ public void init() {
...
     }
 
+
+    private DatastaxCassandraService getCassandraService(Iterable<String> nodes,
+                                                         Integer connectionsPerHostLocal, Integer connectionsPerHostRemote,
+                                                         Integer connectionTimeoutMs, Integer readTimeoutMs) {
+        // this code is borrowed from DatastaxCassandraService::DatastaxCassandraService
+        // but enhances timeout criteria to improve Atlas performance when cluster is loaded
+        PoolingOptions poolingOptions = new PoolingOptions();
+        poolingOptions.setMaxConnectionsPerHost(HostDistance.LOCAL, Preconditions.checkNotNull(connectionsPerHostLocal));
```
emilssolmanis: The methods on `Preconditions` are usually statically imported.

atlas-api/src/main/java/org/atlasapi/AtlasPersistenceModule.java
```diff
@@ -120,6 +126,35 @@ public void init() {
...
+        poolingOptions.setCoreConnectionsPerHost(HostDistance.LOCAL, Preconditions.checkNotNull(connectionsPerHostLocal));
+        poolingOptions.setMaxConnectionsPerHost(HostDistance.REMOTE, Preconditions.checkNotNull(connectionsPerHostRemote));
+        poolingOptions.setCoreConnectionsPerHost(HostDistance.REMOTE, Preconditions.checkNotNull(connectionsPerHostRemote));
+
+        Cluster.Builder builder = Cluster.builder()
+                .addContactPoints((String[]) FluentIterable.from(nodes).toArray(String.class))
+                .withCompression(ProtocolOptions.Compression.SNAPPY)
```
emilssolmanis: Is compression a client-only setting in C* or does it require server config too? If the latter, have we done that config?

cognitive-i: This replication of the existing DatastaxCService.  @Dalamar42 has pointed out that we control DatastaxCassandraService so I'm going to refactor these changes to that project and throw away the replication

PR #194, opened 2015-11-19
Reworking organisation model

Spartakovic:
1.Added datastax organisation store
2.Added kafka worker to listen to changes in organisation store
3.Added legacy organisation resolver
4.Changed legacy transformer
5.Added organisation endpoint


emilssolmanis:
LGTM other than

* we use American spellings throughout the code base, in this case, `Organisation` should be `Organization`.
* you should probably keep an eye on https://github.com/atlasapi/atlas-deer/pull/190 and if it gets merged first, you'll need to add in prep statements here.

atlas-api/src/main/java/org/atlasapi/query/v4/organisation/OrganisationQueryExecutor.java
```diff
@@ -0,0 +1,62 @@
...
+                        throw new UncheckedQueryExecutionException(new NotFoundException(id));
+                    }
+                    Organisation resource = organisationList.get(0);
+                    return QueryResult.singleResult(resource,query.getContext());
+                });
+
+    }
+
+    private ListenableFuture<QueryResult<Organisation>> executeListQuery(final Query<Organisation> query) {
+        throw new UnsupportedOperationException();
```
Dalamar42: How will this exception get mapped to an HTTP response code at the API level? If this generates a 5XX that is not good since this is not a server error, but rather the user requesting an operation that is not supported so we should return something like 400 BAD REQUEST.

Alternatively you can change the request mapping to be '{ "/{id}.*", "/{id}" }' which will guarantee you only get single ID queries.

atlas-api/src/test/java/org/atlasapi/query/v4/organisation/OrganisationQueryExecutorTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+
+        QueryResult<Organisation> queryResult = objectUnderTest.execute(organisationQuery);
+
+        assertThat(queryResult.getOnlyResource(), is(result));
+    }
+
+
+    @Test
+    public void testExecuteMultiFails() throws Exception {
```
Dalamar42: :+1:

atlas-cassandra/src/main/java/org/atlasapi/organisation/DatastaxCassandraOrganisationStore.java
```diff
@@ -0,0 +1,165 @@
...
+    protected DatastaxCassandraOrganisationStore(Session session,
+            ConsistencyLevel writeConsistency, ConsistencyLevel readConsistency) {
+        this.session = checkNotNull(session);
+        this.readConsistency = checkNotNull(readConsistency);
+        this.serializer = new OrganisationSerializer();
+        this.writeConsistency = checkNotNull(writeConsistency);
+
+        RegularStatement statement = select().all()
+                .from(ORGANISATION_TABLE)
+                .where(in(PRIMARY_KEY_COLUMN, bindMarker("keys")));
```
Dalamar42: Don't make the bind markers magic strings. Extract them into constants

atlas-cassandra/src/main/java/org/atlasapi/organisation/DatastaxCassandraOrganisationStore.java
```diff
@@ -0,0 +1,165 @@
...
+
+    @Override public void write(Organisation organisation) {
+        ByteBuffer serializedOrganisation = ByteBuffer.wrap(serializer.serialize(organisation).toByteArray());
+
+        session.execute(rowUpdate.bind()
+                .setLong("organisationId",organisation.getId().longValue())
+                .setBytes("data",serializedOrganisation));
+    }
+
+    public static SessionStep builder() {
```
Dalamar42: Code style: Static methods go after constructors and before non static ones

atlas-cassandra/src/main/java/org/atlasapi/organisation/DatastaxCassandraOrganisationStore.java
```diff
@@ -0,0 +1,165 @@
...
+                    return Resolved.valueOf(
+                            StreamSupport.stream(input.spliterator(), false)
+                                    .map(this::extractOrganisation)
+                                    .collect(Collectors.toList())
+                    );
+                }
+        );
+    }
+
+    protected Organisation extractOrganisation(Row row) {
```
Dalamar42: Code style: Public methods go first (overriden ones should be in the same order as they appear in the interface), then protected, then private

atlas-cassandra/src/main/protobuf/common.proto
```diff
@@ -107,6 +107,7 @@ message Person {
 
 message Organisation {
     repeated Person member = 1;
+    repeated string alternativeTitles = 2;
```
Dalamar42: Does this cover all fields in Organisation?

Dalamar42: Organisation extends ContentGroup so you must make sure you are getting everything from there as well. (BTW that should be done using composition because inheritance is a mess in protocol buffers, i.e. have a content group field. Check event for examples)

PR #195, opened 2015-11-20
Switch to legacy content lister that finds content via the lookup store

Dalamar42:
- The existing implementation could mistakenly return duplicates
where a piece of Content had changed types because Owl orphans rather
than prunes the old version of such Content. Listing via the lookup
store and then resolving via the KnownTypeContentResolver is more
expensive because we have to get a cursor from the lookup store and
then for each entry make one more call to lookup and one to the
appropriate content store, but it's using the same code path as the
Owl API and therefore is more certain to be correct


atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyLookupResolvingContentLister.java
```diff
@@ -0,0 +1,67 @@
+package org.atlasapi.system.legacy;
+
+import static com.google.gson.internal.$Gson$Preconditions.checkNotNull;
```
tfm: *cough*

Dalamar42: Argh... I'm usually careful about that stuff. Fixing...

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyLookupResolvingContentLister.java
```diff
@@ -0,0 +1,67 @@
...
+
+    public LegacyLookupResolvingContentLister(LookupEntryStore lookupEntryStore,
+            KnownTypeContentResolver contentResolver) {
+        this.lookupEntryStore = checkNotNull(lookupEntryStore);
+        this.contentResolver = checkNotNull(contentResolver);
+    }
+
+    /**
+     * This implementation ignores content categories in ContentListingCriteria and only uses
+     * Publishers and ContentListingProgress
```
tfm: Should we throw if categories are provided, then?

Dalamar42: This feels hacky. I think I'll change the lookup store to support categories in the query

PR #197, opened 2015-11-23
Merge batch statements to eliminate possible race conditions

Dalamar42:
- The four batch statements should be safe to merge since only one of
one of them contains particularly large statements
- Delete failing test that is no longer useful. With the deprecation
of the graph column, if an equivalent_content row also has an empty
data column then it has no useful information whatsoever and whether
it resolves or not is immaterial


emilssolmanis:
LGTM :shipit:

PR #198, opened 2015-11-24
Perform a blocking lock rather than tryLock

tfm:
Allow the grouplock to block when obtaining the lock
on the equivalent set, since this is what we're already
doing, functionally.

This is an attempt to simplify the locking logic rather
than any functional change.


tfm:
Superseded.

PR #202, opened 2015-11-24
Feature/cassandra split in statements into multiple id queries

emilssolmanis:


atlas-cassandra/src/main/java/org/atlasapi/content/CassandraEquivalentContentStore.java
```diff
@@ -108,7 +107,7 @@ public CassandraEquivalentContentStore(
 
         RegularStatement statement = select().all()
                 .from(EQUIVALENT_CONTENT_TABLE)
-                .where(eq(SET_ID_KEY, bindMarker("set_id")))
+                .where(eq(SET_ID_KEY, bindMarker("setId")))
```
tfm: `set_id` is part of the clustering key, so this one can remain an `in`

tfm:
:+1:

PR #204, opened 2015-11-25
Split off MetricsRegistry to MetricsModule to eliminate circ dep

Dalamar42:
- Reenable cassandra probe on HealthModule
- Split off MetricsRegistry to MetricsModule
- Delete commented out code from MonitoringWebModule
- Extract AstyanaxProbe to its own class for readability and
separation of concerns. This can't be moved to common-persistence
because we don't have the astyanax driver there

### DIFF
--- a/atlas-processing/src/main/java/org/atlasapi/system/ProcessingMetricsModule.java
+++ b/atlas-processing/src/main/java/org/atlasapi/system/ProcessingMetricsModule.java
@@ -17,7 +17,7 @@ import com.codahale.metrics.jvm.ThreadStatesGaugeSet;
 import com.metabroadcast.common.properties.Configurer;
 
 @Configuration
-public class ProcessingMetricsModule extends MetricsModule {
+public class ProcessingMetricsModule implements MetricsModule {
 
     private final String environmentPrefix = Configurer.get("metrics.environment.prefix").get();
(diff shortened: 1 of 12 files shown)


atlas-processing/src/main/java/org/atlasapi/system/ProcessingMetricsModule.java
```diff
@@ -0,0 +1,50 @@
...
+import com.codahale.metrics.MetricSet;
+import com.codahale.metrics.graphite.GraphiteReporter;
+import com.codahale.metrics.graphite.GraphiteUDP;
+import com.codahale.metrics.jvm.GarbageCollectorMetricSet;
+import com.codahale.metrics.jvm.MemoryUsageGaugeSet;
+import com.codahale.metrics.jvm.ThreadStatesGaugeSet;
+import com.metabroadcast.common.properties.Configurer;
+
+@Configuration
+public class ProcessingMetricsModule extends MetricsModule {
```
emilssolmanis: This doesn't have to inherit from `MetricsModule`. You don't gain anything, you're not using any methods from it, other than just overriding the `@Bean` provider method, and those things should be controlled via `@Qualifier`s and `@Import`. Have you spent too much time staring into the abyss of Fred by any chance?

emilssolmanis:
Other than the inheritance thing, LGTM.

PR #207, opened 2015-11-26
Cache the Dstax session in C*PersistenceModule

emilssolmanis:
to avoid creating multiple sessions (and for the glory of the empire).


Dalamar42:
LGTM

PR #208, opened 2015-11-26
Feature/equivalent content store graph

Dalamar42:
Currently if the equivalent_content store reaches an inconsistent
state where rows exist for a set that should not be part of the set
according to the equivalence graph store we will serve stale content
as part of the set. To resolve this we will store the equivalence
graph as a static field on equivalent_content and use it to check
every returned row is actually part of that set


atlas-cassandra/src/main/java/org/atlasapi/content/CassandraEquivalentContentStore.java
```diff
@@ -404,11 +435,21 @@ private void updateDataColumn(Long setId, Content content) {
...
+                            content.add(deserialized);
+                        }
                     }
                     return content.build();
                 }
-
         );
     }
+
+    private boolean containedInGraph(Id contentId, Row row) {
```
tfm: Pass in an `Optional<EquivalenceGraph>` so we don't deserialize for each check

tfm:
LGTM, other than that minor thing

PR #211, opened 2015-11-28
Implement sorting by multiple fields for ES queries

emilssolmanis:


atlas-core/src/main/java/org/atlasapi/content/QueryOrdering.java
```diff
@@ -1,36 +1,60 @@
...
+            return path;
         }
     }
 
-    public boolean isAscending() {
-        return ascending;
+    private final ImmutableList<Node> sortOrder;
+
+    public QueryOrdering(ImmutableList<Node> sortOrder) {
+        this.sortOrder = sortOrder;
```
tfm: `checkNotNull`?

tfm:
back to you, @emilssolmanis

atlas-core/src/main/java/org/atlasapi/content/QueryOrdering.java
emilssolmanis: Oh bloody hell, we should just have a checkstyle rule for this or something...

tfm: :)

atlas-core/src/main/java/org/atlasapi/content/QueryOrdering.java
```diff
@@ -2,35 +2,61 @@
...
+    public enum Direction {
+        ASC,
+        DESC
     }
 
-    public static QueryOrdering fromOrderBy(String orderBy) {
-        int lastDot = orderBy.lastIndexOf(".");
-        if (lastDot == -1) {
-            throw new IllegalArgumentException("Missing .asc or .desc operator after " + orderBy);
+    public static class Node {
```
cognitive-i: Node doesn't feel like the right name here.  Maybe Clause/Term would be better?

atlas-elasticsearch/src/main/java/org/atlasapi/content/EsUnequivalentContentIndex.java
```diff
@@ -233,15 +233,17 @@ private void addActionableFilter(Optional<IndexQueryParams> queryParams,
...
-        } else {
-            reqBuilder.addSort(
-                    SortBuilders
-                            .fieldSort(translateOrderField(order.getPath()))
-                            .missing("_last")
-                            .order(order.isAscending() ? SortOrder.ASC : SortOrder.DESC)
-            );
+        for (QueryOrdering.Node node : order.getSortOrder()) {
+            if ("relevance".equalsIgnoreCase(node.getPath())) {
+                reqBuilder.addSort(SortBuilders.scoreSort().order(SortOrder.DESC));
```
cognitive-i: why is this special case required?  Surely if the user specified something silly they should be notified early rather than cover-up?

atlas-elasticsearch/src/test/java/org/atlasapi/content/EsUnequivalentContentIndexTest.java
```diff
@@ -386,20 +396,142 @@ public void testQueryOrder() throws Exception {
...
+        );
 
         AttributeQuerySet querySet = new AttributeQuerySet(ImmutableList.of(query));
-        ListenableFuture<IndexQueryResult> result = index.query(querySet,
+        ListenableFuture<IndexQueryResult> result = index.query(
+                querySet,
                 ImmutableList.of(Publisher.METABROADCAST),
-                Selection.all(), Optional.empty());
+                Selection.all(),
+                Optional.of(new IndexQueryParams(
```
cognitive-i: A comment would be helpful here to describe intent of such a strange construction

cognitive-i:
Looks pretty good to me, just a few extra comments and naming of Node would make it extra good

atlas-elasticsearch/src/main/java/org/atlasapi/content/EsUnequivalentContentIndex.java
emilssolmanis: @cognitive-i because we need a way to expose ElasticSearch's search relevance score to the public API, in this case we do it as `relevance`.

atlas-elasticsearch/src/test/java/org/atlasapi/content/EsUnequivalentContentIndexTest.java
emilssolmanis: @cognitive-i I'd argue that it's clear from the test name, which is what it should be.

PR #214, opened 2015-12-01
Merge broadcasts where start times are slightly different

tfm:
We have a tolerance in equivalence, but did not have one for
broadcast merging on output. We will now match broadcasts
between equivalent items if they have a start date within
five minutes of each other.


Dalamar42:
LGTM

PR #221, opened 2015-12-04
Clamp location availability start/end to deal with Year 2038 bug

Dalamar42:
- This attempts to fix the Unix Millennium Bug for 32-bit clients by
changing any date time that is greater than 2038-01-01T00:00:00.0 to
that value
- This was needed because some Android devices are 32-bit and we
ingest from content providers that sometimes give us dates in the
year 3000
- This "fix" is set to be disabled automatically on
2016-03-05T08:00:00.0Z and there is a unit test set to fail then to
alert us to remove this


emilssolmanis:
Assuming you're using it in all the necessary places, :shipit:

PR #226, opened 2015-12-10
Adding new fields (vat is Double, requiredEncryption is boolean, subtitledLanguages is List<String>) to location

timothy22000:
- vat is Double, requiredEncryption is boolean, subtitledLanguages is List<String>
- Added 3 new fields to Location in Deer
- Updated Content.Proto with the 3 new fields.
- Updated LegacyContentTransformer to handle the new fields
- Updated LocationSerializer to serialize and deserialize the new fields.

### DIFF
--- a/atlas-cassandra/src/main/java/org/atlasapi/content/LocationSerializer.java
+++ b/atlas-cassandra/src/main/java/org/atlasapi/content/LocationSerializer.java
@@ -11,6 +11,7 @@ import org.atlasapi.serialization.protobuf.ContentProtos;
 import org.atlasapi.serialization.protobuf.ContentProtos.Location.Builder;
 import org.atlasapi.util.ImmutableCollectors;
 
+import com.google.common.base.Strings;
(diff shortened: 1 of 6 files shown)


atlas-cassandra/src/main/java/org/atlasapi/content/LocationSerializer.java
```diff
@@ -33,6 +34,19 @@
             builder.setTransportType(location.getTransportType().toString());
         }
         if (location.getUri() != null) { builder.setUri(location.getUri()); }
+
+        if(location.getVat() != null) { builder.setVat(location.getVat()); }
```
Dalamar42: Always put if blocks in multiple lines as per code style

Dalamar42:
Generally the code style dictates spaces between say an if an paren. So:
if () instead of if(). Same for the others.

Dalamar42:
Some code style comments. Otherwise LGTM

PR #230, opened 2015-12-15
Fix deadlock in `AbstractEquivalenceGraphStore`.

emilssolmanis:
When updating equivalences, if a corrupted graph was detected, the graph store would clear it and retry the update. The retry was being done via recursion and the lock used is not reentrant, which causes a deadlock. This hacks around it by unlocking before recursing.

### DIFF
--- a/atlas-core/src/main/java/org/atlasapi/equivalence/AbstractEquivalenceGraphStore.java
+++ b/atlas-core/src/main/java/org/atlasapi/equivalence/AbstractEquivalenceGraphStore.java
@@ -97,19 +97,37 @@ public abstract class AbstractEquivalenceGraphStore implements EquivalenceGraphS
             throw new WriteException(e);
         } catch (CorruptEquivalenceGraphIndexException e) {
             cleanGraphAndIndex(e.getSubjectId());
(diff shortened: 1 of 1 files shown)


atlas-core/src/main/java/org/atlasapi/equivalence/AbstractEquivalenceGraphStore.java
```diff
@@ -91,25 +91,44 @@ public AbstractEquivalenceGraphStore(MessageSender<EquivalenceGraphUpdateMessage
             
         } catch(InterruptedException e) {
             log().error(String.format("%s: %s", subject, newAdjacents), e);
+            unlock(subjectAndAdjacents, transitiveSetsIds);
```
Dalamar42: You don't need to call unlock() here. The finally block will be executed before the return

Dalamar42: Gagh, just noticed which exception you are catching. Ignore me

Dalamar42:
LGTM

PR #231, opened 2016-01-04
Change users of MessageSender to specify partition key

Dalamar42:
- This is intended to ensure messages about the same resource always
go the same partition and that therefore we have ordering guarantees
about those messages. This is irrelevant when it comes to idempotent
messages, but it is required to ensure correctness when it comes to
non-idempotent ones like EquivalenceGraphUpdateMessage


emilssolmanis:
LGTM :shipit:

PR #232, opened 2016-01-06
Api key from header

Spartakovic:
https://jira.metabroadcast.com/browse/MBST-13668
This will allow users to set an api key in a header


Dalamar42:
LGTM

PR #235, opened 2016-01-08
Sort ES hits according to publisher precedence using groovy script

Dalamar42:
- Currently due to an issue we can't inline groovy scripts to
requests made to the production cluster so we have to operate in two
modes. We inline the script for integration tests and reference a
script saved on the servers for production
- When ordering is requested and there is no fuzzy search which
requires score sorting we will also sort according to publisher
precedence to ensure that when deduping we are more likely to sort
according the value that will be displayed to the user


tfm:
:+1:

PR #237, opened 2016-01-11
Improve logging for equivalence graph and equiv content graph workers

Dalamar42:


atlas-processing/src/main/java/org/atlasapi/messaging/ContentEquivalenceUpdatingWorker.java
```diff
@@ -43,15 +43,16 @@ public ContentEquivalenceUpdatingWorker(EquivalenceGraphStore graphStore, @Nulla
 
     @Override
     public void process(EquivalenceAssertionMessage message) throws RecoverableException {
-        LOG.debug("Processing message on id {}, message: {}", message.getSubject().getId(),
-                message);
+        LOG.debug(
+                "Processing message on id {}, asserted adjacents: {}, message: {}",
+                message.getSubject().getId(),
+                message.getAssertedAdjacents().stream()
```
emilssolmanis: This being evaluated on each message might warrant having this within a `if (LOG.isDebugEnabled())` block.

emilssolmanis:
Looks sane, other than maybe wrapping the streams in checks, since they get evaluated for every message and we probably don't want that.

PR #245, opened 2016-01-14
Bug/missed graph creation handling

Dalamar42:
- When deleting a graph in the equivalent content store check all
content belonging to that set. If that content cannot be found in
the updated or created graphs that are part of the same update
message then assume that could be stale content that was caused by
a previous missed update and update it immediately to avoid 404'ing
when someone requests those content IDs via the API
- When a single piece of content is updated in the equivalent content
store also update the corresponding entry in the equivalent content
index to ensure we maximise the chances of fixing stale entries in
the index caused by past messages that did not get processed correctly
- The updateContent method in the AbstractEquivalentContentStore
should accept content ID instead of the content itself and only
resolve the content after it has acquired a lock on the content ID.
This avoids race conditions when there are concurrent updates on the
same content


emilssolmanis:
:shipit:

PR #249, opened 2016-01-15
Change ES deduping to take duplicate with highest precedence publisher

Dalamar42:
- Currently we grab the ID we see out of every equivalent set. This
creates the problem that after merging on output we might show
different values on the API than the ones used for ordering by ES
thus the ordering will appear wrong. This is intended to fix that
- Change IndexQueryResult to only store IDs and create separate
DelegateIndexQueryResult to store additional information required by
the PseudoEquivalentContentIndex (namely canonicalID and publisher) to
avoid leaking that information to other parts of the code that don't
need them and making those interfaces more confusing. The interface
DelegateContentIndex has been created to handle communication between
the PseudoEquivalentContentIndex and EsUnequivalentContentIndex
- Have PseudoEquivalentContentIndex return the normal ID of the
content chosen after deduping. This is better than returning the
canonicalId since--deduping aside--the ES store does not store
equivalent sets and therefore should not be returning a canonical ID.
It also eliminates several issues we've had caused by inconsistencies
of canonical ID between the ES store and the equivalent content store

### DIFF
--- a/atlas-elasticsearch/src/main/java/org/atlasapi/content/DelegateContentIndex.java
+++ b/atlas-elasticsearch/src/main/java/org/atlasapi/content/DelegateContentIndex.java
@@ -1,0 +1,22 @@
+package org.atlasapi.content;
+
+import java.util.Optional;
+
+import org.atlasapi.criteria.AttributeQuerySet;
+import org.atlasapi.media.entity.Publisher;
+
+import com.metabroadcast.common.query.Selection;
+
+import com.google.common.util.concurrent.ListenableFuture;
+
+/**
+ * This interface is used when a content index is used as a delegate of another content index to
+ * pass additional information between them without polluting the {@link ContentIndex} interface
+ * with it.
+ */
+public interface DelegateContentIndex {
+
+    ListenableFuture<DelegateIndexQueryResult> delegateQuery(AttributeQuerySet query,
+            Iterable<Publisher> publishers, Selection selection,
+            Optional<IndexQueryParams> queryParams);
+}
--- a/atlas-elasticsearch/src/main/java/org/atlasapi/content/EsUnequivalentContentIndex.java
+++ b/atlas-elasticsearch/src/main/java/org/atlasapi/content/EsUnequivalentContentIndex.java
@@ -1,7 +1,5 @@
 package org.atlasapi.content;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-
 import java.io.IOException;
 import java.util.Optional;
 import java.util.stream.StreamSupport;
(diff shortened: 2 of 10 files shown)


atlas-elasticsearch/src/main/java/org/atlasapi/content/DelegateContentIndex.java
```diff
@@ -0,0 +1,17 @@
...
+import java.util.Optional;
+
+import org.atlasapi.criteria.AttributeQuerySet;
+import org.atlasapi.media.entity.Publisher;
+
+import com.metabroadcast.common.query.Selection;
+
+import com.google.common.util.concurrent.ListenableFuture;
+
+public interface DelegateContentIndex {
```
emilssolmanis: It'd be nice to have at least some minor docs on why this exists, since it's definitely not obvious.

atlas-elasticsearch/src/main/java/org/atlasapi/content/EsUnequivalentContentIndex.java
```diff
@@ -274,4 +304,14 @@ private Id getCanonicalId(SearchHit hit) {
         Long id = hit.field(EsContent.CANONICAL_ID).<Number>value().longValue();
         return Id.valueOf(id);
     }
+
+    private Publisher getSource(SearchHit hit) {
```
emilssolmanis: While it's `private`, this should probably still return an `Optional`.

atlas-elasticsearch/src/main/java/org/atlasapi/content/PseudoEquivalentContentIndex.java
```diff
@@ -83,55 +76,67 @@ private Selection getSelectionForDelegate(Iterable<Publisher> publishers, Select
...
+            Iterable<Publisher> publishers) {
+        LinkedHashMap<Id, DelegateIndexQueryResult.Result> dedupedEntries = new LinkedHashMap<>();
 
-    private ImmutableList<Id> paginateIds(Iterable<Id> ids, Selection selection) {
-         return StreamSupport.stream(ids.spliterator(), false)
-                .skip(selection.hasNonZeroOffset() ? selection.getOffset() : 0)
-                .limit(selection.limitOrDefaultValue(100))
-                .collect(ImmutableCollectors.toList());
-    }
+        for (DelegateIndexQueryResult.Result entry : result.getResults()) {
```
emilssolmanis: :hankey:

atlas-elasticsearch/src/main/java/org/atlasapi/content/PseudoEquivalentContentIndex.java
```diff
@@ -83,55 +76,67 @@ private Selection getSelectionForDelegate(Iterable<Publisher> publishers, Select
...
+                return true;
             }
-            LOG.warn("Found no canonical ID for {} using {}", id, id);
-        } catch (IOException e) {
-            LOG.warn("Found no canonical ID for {} using {}", id, id, e);
         }
-        return Optional.empty();
+
+        // This should never happen as the delegate uses the same publishers for the query
+        LOG.warn("Delegate content index returned publisher {} that was not asked for."
```
emilssolmanis: If this should really **never** __ever__ happen, we should probably log it as an error, not warning.

emilssolmanis:
Other than the minor nitpicks, lgtm :shipit:

PR #266, opened 2016-01-19
Fix bug in actionable content filter

Dalamar42:
- Split failing actionable content filter test into multiple tests
that each cover a single scenario to assess which scenario is failing
- Fix actionable filter bug where no region ID has been provided. In
that scenario the filters were written in the wrong order (nesting
outside the or filter rather than the other way around)


emilssolmanis:
:+1:

PR #267, opened 2016-01-20
Fix NPE when transforming legacy images with null canonical URI

Dalamar42:
- Drop images with null canonical URI when transforming legacy images
to the Deer model


atlas-legacy/src/main/java/org/atlasapi/system/legacy/BaseLegacyResourceTransformer.java
```diff
@@ -41,36 +41,33 @@
...
 		}
 
-		return Iterables.transform(images, new Function<org.atlasapi.media.entity.Image, Image>() {
-			@Override
-			public Image apply(org.atlasapi.media.entity.Image input) {
-				return transformImage(input);
-			}
-		});
+        return images.stream()
+                .filter(image -> image.getCanonicalUri() != null)
```
tfm: Perhaps add a comment as to why we're doing this

PR #270, opened 2016-01-25
Endpoint that deactivates a users third-party (Google/Github/Twitter) account from atlas admin

Spartakovic:
-Updated User model with new field
-Added endpoint to set and store new field


MartinsIrbe:
Tests?

MartinsIrbe:
lgtm

atlas-api/src/main/java/org/atlasapi/application/users/UsersController.java
```diff
@@ -146,6 +152,38 @@ public void updateUser(HttpServletRequest request,
             new ErrorResultWriter().write(summary, writer, request, response);
         }
     }
+
+    @RequestMapping(value = "/4/users/deactivate/{uid}.*", method = RequestMethod.POST)
```
MartinsIrbe: Would be better -  4/users/{uid}.*/deactivate

atlas-api/src/main/java/org/atlasapi/application/users/UsersController.java
```diff
@@ -146,6 +152,38 @@ public void updateUser(HttpServletRequest request,
             new ErrorResultWriter().write(summary, writer, request, response);
         }
     }
+
+    @RequestMapping(value = "/4/users/{uid}/deactivate.*", method = RequestMethod.POST)
```
emilssolmanis: We've agreed to not use verbs in paths, so this needs to be fixed.

Spartakovic: I am just updating one field, but don't delete anything, does it make any difference? I am not sure of the semantics of "DELETE", I thought it is applied only when we actually delete something. I can update it

emilssolmanis: @Spartakovic it's the outside semantics that matter, not the underlying data model. @Dalamar42 @tfm what do you think?

PR #271, opened 2016-01-25
Added generic enum to Image.Type and updated ImageListWriter to not s…

timothy22000:
…et image if it is generic

### DIFF
--- a/atlas-core/src/main/java/org/atlasapi/content/Image.java
+++ b/atlas-core/src/main/java/org/atlasapi/content/Image.java
@@ -75,7 +75,8 @@ public class Image implements Sourced {
         ADDITIONAL("additional"),
         BOX_ART("box_art"),
         POSTER("poster"),
-        LOGO("logo");
+        LOGO("logo"),
+        GENERIC("generic");
         
         private final String name;
         


atlas-api/src/main/java/org/atlasapi/output/writers/ImageListWriter.java
```diff
@@ -20,7 +20,7 @@ public void write(Image entity, FieldWriter writer, OutputContext ctxt) throws I
         writer.writeField("mime_type", entity.getMimeType());
         
         Image.Type type = entity.getType();
-        writer.writeField("type", type != null ? entity.getType().getName() : null);
+        writer.writeField("type", (type != null || type != Image.Type.GENERIC) ? entity.getType().getName() : null);
```
tfm: Why do we need to special-case `Image.Type.GENERIC` here? It's just another value in the enum isn't it?

tfm:
Back to you, @timothy22000

atlas-api/src/main/java/org/atlasapi/output/writers/ImageListWriter.java
timothy22000: It was for this ticket. https://jira.metabroadcast.com/browse/MBST-13789

I made a similar change in Owl ImageSimplifer. The special case is to only set images if it is not generic. 

Did I misinterpret the ticket?

PR #272, opened 2016-01-26
added annotation to return non-merged content

Spartakovic:
The Data Comparison Service needs to be able to see unequivalated output, so that each data source's data can be rendered and compared by a user.


emilssolmanis:
lgtm :shipit:

PR #273, opened 2016-01-27
Do not resolve equivalence graphs from stale index entries

Dalamar42:
- Due to past bugs we could have stale index entries that point to an
equivalence graph that does not actually contain the resource ID in
question. This will prevent resolution of the graph from the stale
index and will instead return absent, i.e. that a graph has not been
found for the requested resource ID
- Also this removes the code in AbstractEquivalenceGraphStore that
was attempting to handle this issue by deleting the "corrupt" graph
because the graph that is resolved from the stale index is not
necessarily a bad one. At that point all we know is that it should
not have been resolved for that particular requested resource ID

### DIFF
--- a/atlas-cassandra/src/main/java/org/atlasapi/equivalence/CassandraEquivalenceGraphStore.java
+++ b/atlas-cassandra/src/main/java/org/atlasapi/equivalence/CassandraEquivalenceGraphStore.java
@@ -1,13 +1,5 @@
 package org.atlasapi.equivalence;
 
-import static com.datastax.driver.core.querybuilder.QueryBuilder.bindMarker;
-import static com.datastax.driver.core.querybuilder.QueryBuilder.delete;
-import static com.datastax.driver.core.querybuilder.QueryBuilder.eq;
-import static com.datastax.driver.core.querybuilder.QueryBuilder.insertInto;
-import static com.datastax.driver.core.querybuilder.QueryBuilder.select;
-import static com.google.common.base.Preconditions.checkNotNull;
-
-import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


atlas-cassandra/src/main/java/org/atlasapi/equivalence/CassandraEquivalenceGraphStore.java
```diff
@@ -113,24 +100,7 @@ public CassandraEquivalenceGraphStore(MessageSender<EquivalenceGraphUpdateMessag
...
-            /* Remove graph entry */
-            log.warn("Deleting graph {} for subject {}", graphId, subjectId);
-            session.execute(graphDelete.bind(graphId));
-        } catch (IOException e) {
-            throw new RuntimeException(e);
-        }
-    }
-
-    private final Function<Iterable<Row>, Map<Long, EquivalenceGraph>> toGraph = rows -> {
+    private Function<Iterable<Row>, Map<Long, EquivalenceGraph>> toGraph = rows -> {
```
emilssolmanis: Why remove `final`?

atlas-cassandra/src/main/java/org/atlasapi/equivalence/CassandraEquivalenceGraphStore.java
```diff
@@ -139,7 +109,7 @@ protected void cleanGraphAndIndex(Id subjectId) {
             return idGraph.build();
         };
 
-    private final AsyncFunction<Map<Id, Long>, OptionalMap<Id, EquivalenceGraph>> toGraphs = idIndex -> {
+    private AsyncFunction<Map<Id, Long>, OptionalMap<Id, EquivalenceGraph>> toGraphs = idIndex -> {
```
emilssolmanis: Same, why remove `final`?

atlas-core/src/test/java/org/atlasapi/equivalence/AbstractEquivalenceGraphStoreTest.java
```diff
@@ -638,13 +673,12 @@ protected Logger log() {
...
-        protected void cleanGraphAndIndex(Id subjectId) {
-            store.remove(subjectId);
-        }
-
-        @Override
         protected GroupLock<Id> lock() {
             return lock;
         }
+
+        protected void cleanGraphAndIndex(Id subjectId) {
```
emilssolmanis: This is unnecessary now, you removed this from the abstract base.

emilssolmanis:
Other than the minor bits, :shipit:

PR #277, opened 2016-02-01
Fail astyanax probe on no active hosts

emilssolmanis:


atlas-processing/src/main/java/org/atlasapi/system/health/AstyanaxProbe.java
```diff
@@ -1,13 +1,14 @@
...
-import static com.google.gson.internal.$Gson$Preconditions.checkNotNull;
-
 import com.metabroadcast.common.health.HealthProbe;
 import com.metabroadcast.common.health.ProbeResult;
+
 import com.netflix.astyanax.AstyanaxContext;
 import com.netflix.astyanax.Keyspace;
 import com.netflix.astyanax.connectionpool.ConnectionPoolMonitor;
 
+import static com.google.gson.internal.$Gson$Preconditions.checkNotNull;
```
tfm: *cough*

tfm:
:+1:

PR #279, opened 2016-02-02
Returning non-merged content

Spartakovic:
1.Added new annotation
2.With the new annotation content is not being merged
3.Fix passing nulls

### DIFF
--- a/atlas-api/src/main/java/org/atlasapi/query/v4/content/IndexBackedEquivalentContentQueryExecutor.java
+++ b/atlas-api/src/main/java/org/atlasapi/query/v4/content/IndexBackedEquivalentContentQueryExecutor.java
@@ -73,12 +73,8 @@ public class IndexBackedEquivalentContentQueryExecutor implements QueryExecutor<
                     if (equivs.isEmpty()) {
                         throw new UncheckedQueryExecutionException(new NotFoundException(contentId));
                     }
-                    if (equivs.size() == 1) {
-                        Content resource = equivs.get(0);
-                        return QueryResult.singleResult(resource, query.getContext());
-                    } else {
-                        return QueryResult.listResult(equivs, query.getContext(), equivs.size());
-                    }
+                    Content resource = equivs.get(0);
+                    return QueryResult.singleResult(resource, query.getContext());
                 }
         );
     }
--- a/atlas-core/src/main/java/org/atlasapi/content/MergingEquivalentsResolverBackedContainerSummaryResolver.java
+++ b/atlas-core/src/main/java/org/atlasapi/content/MergingEquivalentsResolverBackedContainerSummaryResolver.java
@@ -29,11 +29,11 @@ public class MergingEquivalentsResolverBackedContainerSummaryResolver implements
     }
 
     @Override
-    public Optional<ContainerSummary> resolveContainerSummary(Id id, ApplicationSources applicationSources) {
+    public Optional<ContainerSummary> resolveContainerSummary(Id id, ApplicationSources applicationSources, Set<Annotation> annotations) {
         ResolvedEquivalents<Content> contentResolved = null;
         try {
             contentResolved = Futures.get(
-                    contentResolver.resolveIds(ImmutableSet.of(id), applicationSources, ImmutableSet.<Annotation>of()),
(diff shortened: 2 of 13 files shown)


atlas-api/src/main/java/org/atlasapi/query/v4/content/IndexBackedEquivalentContentQueryExecutor.java
```diff
@@ -73,8 +73,12 @@ public IndexBackedEquivalentContentQueryExecutor(ContentIndex contentIndex,
                     if (equivs.isEmpty()) {
                         throw new UncheckedQueryExecutionException(new NotFoundException(contentId));
                     }
-                    Content resource = equivs.get(0);
-                    return QueryResult.singleResult(resource, query.getContext());
+                    if (equivs.size() == 1) {
```
tfm: Why has this additional check been added?

atlas-core/src/main/java/org/atlasapi/content/MergingEquivalentsResolverBackedContainerSummaryResolver.java
```diff
@@ -31,7 +33,7 @@ public MergingEquivalentsResolverBackedContainerSummaryResolver(MergingEquivalen
         ResolvedEquivalents<Content> contentResolved = null;
         try {
             contentResolved = Futures.get(
-                    contentResolver.resolveIds(ImmutableSet.of(id), applicationSources, null),
+                    contentResolver.resolveIds(ImmutableSet.of(id), applicationSources, ImmutableSet.<Annotation>of()),
```
tfm: Do we not have any annotations here?

atlas-api/src/main/java/org/atlasapi/query/v4/content/IndexBackedEquivalentContentQueryExecutor.java
Spartakovic: Because I added new annotation that returns set of non-merged equvalated items. Before it was returning just one merged item

atlas-core/src/main/java/org/atlasapi/content/MergingEquivalentsResolverBackedContainerSummaryResolver.java
Spartakovic: Null was being passed here, so I assumed that no annotations were needed

Spartakovic: Also the only time when annotation is used - the annotation I just added, but since container summary should be merged, I think no annotation is needed

atlas-api/src/main/java/org/atlasapi/query/v4/content/IndexBackedEquivalentContentQueryExecutor.java
tfm: Are we expecting a single item to be returned, or a list of equivalent items? I'd expect just the resource that was requested.

atlas-core/src/main/java/org/atlasapi/equivalence/AnnotationBasedMergingEquivalentsResolver.java
```diff
@@ -29,10 +29,14 @@ public DefaultMergingEquivalentsResolver(EquivalentsResolver<E> resolver, Applic
     @Override
     public ListenableFuture<ResolvedEquivalents<E>> resolveIds(Iterable<Id> ids,
             ApplicationSources sources, Set<Annotation> activeAnnotations) {
-        ListenableFuture<ResolvedEquivalents<E>> unmerged
-            = resolver.resolveIds(ids, sources.getEnabledReadSources(), activeAnnotations);
-        
-        return Futures.transform(unmerged, mergeUsing(sources));
+
+        if(!activeAnnotations.contains(Annotation.NON_MERGED)) {
```
Dalamar42: Best avoid negative conditions where it's reasonable. Positive assertions are a bit easier to reason about and that "!" can easily be missed. Here you can just flip the if.

Dalamar42:
One small comment. Otherwise LGMT

Dalamar42:
+ Tests :)

PR #281, opened 2016-02-02
Use EPoll transport for Netty.

emilssolmanis:
This should reduce GC churn from various `byte[]` allocations in the C* driver.

Tested on `stage`, basic heap profiling says it's good at getting rid of the crazy amounts of `byte[]` allocations we have, and GC gets a bit better. As far as I can tell, there are no downsides.


tfm:
:+1:

Dalamar42:
LGMT

PR #285, opened 2016-02-05
Parse the schedule override argument and support it on the schedule endpoint

emilssolmanis:


atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleRequestParser.java
```diff
@@ -120,14 +135,31 @@ private ScheduleQuery parseSingleRequest(HttpServletRequest request)
...
 
         DateTime from = extractFrom(request);
+        Optional<Publisher> override = extractOverride(request);
+
         Optional<DateTime> to = extractTo(request);
         if (to.isPresent()) {
             checkInterval(from, to.get());
-            return ScheduleQuery.single(publisher, from, to.get(), context, channel);
+            return ScheduleQuery.single(
+                    publisher, override.orElse(null),
```
Dalamar42: I presume the null is due to an existing method that expects null?

atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleQuery.java
```diff
@@ -36,23 +51,52 @@ public static final ScheduleQuery multi(Publisher source, DateTime start, Intege
...
     private final Optional<Integer> count;
     private final QueryContext context;
+    private final Optional<Publisher> override;
+
+    public ScheduleQuery(
+            Publisher source, Publisher override,
+            DateTime start, DateTime end, QueryContext context) {
+        this(
+                checkNotNull(source),
+                Optional.fromNullable(override),
```
Dalamar42: Why are you not using overloading or just passing Optional here instead of null?

emilssolmanis: Cause I was lazy and there'd be 4 permutations instead of 2. I know, I know, will fix.

PR #287, opened 2016-02-10
Be less eager to remove stale broadcasts

tfm:
In the case where a broadcast ID is used on mulitple broadcasts
across two days, and one is removed, if the update interval of
the schedule spans both days, then all broadcasts with that
ID were removed.

We now only remove broadcasts on the calendar day in which the
stale broadcast is. There remains an edge case if both the valid
and invalid broadcast are on the same day.


Dalamar42:
LGTM

PR #288, opened 2016-02-11
New EventStore to cope with organisationRefs (old eventstore will be decomissioned, once everything is done)

Spartakovic:
- Added new Event model to have OrganisationRefs instead of organisations
- Added new event related classes: store, hasher
- Added Organisation Ref serializer
- Added tests
- Added new event protobuf

### DIFF
--- a/atlas-cassandra/src/main/java/org/atlasapi/eventV2/DatastaxCassandraEventStoreV2.java
+++ b/atlas-cassandra/src/main/java/org/atlasapi/eventV2/DatastaxCassandraEventStoreV2.java
@@ -1,0 +1,221 @@
+package org.atlasapi.eventV2;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import org.atlasapi.content.CorruptContentException;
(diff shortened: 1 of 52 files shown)


atlas-cassandra/src/main/java/org/atlasapi/eventV2/DatastaxCassandraEventStoreV2.java
```diff
@@ -0,0 +1,221 @@
...
+
+public class DatastaxCassandraEventStoreV2 implements EventV2PersistenceStore {
+
+    private static final String EVENT_TABLE = "event_v2";
+    private static final String PRIMARY_KEY_COLUMN = "event_id";
+    private static final int TIMEOUT_IN_MINUTES = 1;
+
+    private final Logger log = LoggerFactory.getLogger(getClass());
+
+    private final AliasIndex<EventV2> aliasIndex;
```
Dalamar42: I'm not sure we need the alias store, but keep it around for now.
Though make sure you pass in a *different* (i.e. v2) alias store than the one the other Event store is using to avoid any potential conflicts.

Dalamar42: PS: That means a new C* event aliases table schema as well

Dalamar42:
Aside from the one comment LGTM

PR #289, opened 2016-02-11
Prevent deadlock in AbstractEquivalentContentStore

Dalamar42:
- The lock used in AbstractEquivalentContentStore is not reentrant
which can cause a deadlock when updating stale content during a graph
deletion. The ID(s) of the deleted graph(s) will be in the locked
IDs, but could also be among the stale content IDs which will cause
the thread to block itself when it tries to lock them before
updating stale content


tfm:
:+1:

PR #290, opened 2016-02-15
Feature/schedule overrides

emilssolmanis:
This looks scary, but it only really does 3 things:

* adds a `ScheduleMerger` interface and a (rather complicated) implementation. You might twist your brain trying to comprehend it, but it's 100% covered by tests for any possible edge case I could think\
 of. Suggestions for how to rewrite this better sorely needed, current consensus seems that we might be able to rewrite this with a 2-pass thing.
* adds an "override_sources" query arg to `ScheduleController` + implements relevant things in the parser and query object
* adds calls to the `ScheduleMerger` in the query executor

### DIFF
--- a/atlas-api/src/main/java/org/atlasapi/query/v4/schedule/EquivalentScheduleQueryExecutor.java
+++ b/atlas-api/src/main/java/org/atlasapi/query/v4/schedule/EquivalentScheduleQueryExecutor.java
@@ -2,6 +2,7 @@ package org.atlasapi.query.v4.schedule;
 
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 import org.atlasapi.application.ApplicationSources;
 import org.atlasapi.channel.Channel;
@@ -26,6 +27,7 @@ import org.atlasapi.schedule.EquivalentScheduleResolver;
 import org.atlasapi.schedule.FlexibleBroadcastMatcher;
 import org.atlasapi.schedule.Schedule;
 
+import com.codepoetics.protonpack.StreamUtils;
 import com.google.common.base.Function;
 import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableList;
@@ -47,6 +49,7 @@ public class EquivalentScheduleQueryExecutor
     private EquivalentScheduleResolver scheduleResolver;
     private ApplicationEquivalentsMerger<Content> equivalentsMerger;
     private FlexibleBroadcastMatcher broadcastMatcher;
+    private ScheduleMerger scheduleMerger = new ScheduleMergerImpl();
 
     public EquivalentScheduleQueryExecutor(
             ChannelResolver channelResolver,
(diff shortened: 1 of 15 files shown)


atlas-api/src/main/java/org/atlasapi/query/v4/schedule/EquivalentScheduleQueryExecutor.java
```diff
@@ -83,20 +96,49 @@ public EquivalentScheduleResolverBackedScheduleQueryExecutor(
...
         if (query.isMultiChannel()) {
-            List<ChannelSchedule> channelSchedules = channelSchedules(schedule, query);
+            List<ChannelSchedule> channelSchedules;
+
+            if (query.getOverride().isPresent()) {
+                List<ChannelSchedule> originalSchedules = channelSchedules(schedule, query);
+                List<ChannelSchedule> overrideSchedules = channelSchedules(overrideSchedule, query);
+
+                channelSchedules = StreamUtils
+                        .zip(originalSchedules.stream(), overrideSchedules.stream(),
```
Dalamar42: Presumably the `originalSchedules` and the `overrideSchedules` will have the exact same schedules, in the same order and with no gaps, yes?

atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleMergerImpl.java
```diff
@@ -0,0 +1,229 @@
...
+                            originalSchedule.getInterval().getEnd()
+                    ),
+                    mergedBuilder.build()
+            );
+        }
+
+        ItemAndBroadcast curr;
+        ItemAndBroadcast override = null;
+
+        boolean removeFollowOn = false;
```
Dalamar42: Maybe set these as static finals rather than set them in the method itself

emilssolmanis: Huh?

Dalamar42: Actually nm. I think I misunderstood something. These are not flags to enable behaviour, but are rather tracking state during the merge?

atlas-api/src/main/java/org/atlasapi/query/v4/schedule/EquivalentScheduleQueryExecutor.java
emilssolmanis: I should hope so. I should also probably throw if they don't.

atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleMergerImpl.java
```diff
@@ -0,0 +1,229 @@
...
+            removeFollowOn = false;
+            truncateFollowOnToOverride = false;
+
+            if (override == null) {
+                override = itOverride.next();
+            }
+
+            Broadcast overrideBroadcast = override.getBroadcast();
+
+            /* currStart >= overrideEnd, disjoint strictly after
```
Dalamar42: Missing 'if'

PR #292, opened 2016-02-15
New Event Annotation hydrating organisation refs

Spartakovic:


Dalamar42:
You haven't wired the annotation in the API. I presume you will do that when you are ready to switch the API to the new store?

Otherwise LGTM

PR #294, opened 2016-02-16
MBST-13893: Logging: switch to log4j2 with slf4j bridge

cognitive-i:
* intent is to log to Kafka via log4j2-logstash-jsonevent-layout
* kafka logging is configured via puppet
* rudimentary fileroller configs have replaced Atlas-API and Atlas-Cassandra


atlas-api/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,41 @@
...
+
+    </RollingFile>
+  </Appenders>
+
+  <Loggers>
+    <Root level="info">
+      <AppenderRef ref="RollingFile"/>
+    </Root>
+
+    <Logger name="org.atlasapi.persistence.content.mongo" level="debug">
```
Dalamar42: I don't think we should have these custom logger levels here. For log4j we had two files `log4j.core` and `log4j.extension` which are then merged to `log4j.properties` by the `jetty` script. The `log4j.extension` file is not being managed by puppet so we put temporary trace, debug entries there.

FYI the `jetty` just does a cat so this won't work with XML. Can we use properties here instead?

atlas-api/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration status="warn" name="Atlas-API" packages="*">
+
+  <Appenders>
+    <RollingFile name="RollingFile" fileName="atlas-api-test.log">
```
Dalamar42: Do we really want a file appender for tests? Why not just stdout?

atlas-cassandra/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration status="warn" name="Atlas-Cassandra" packages="*">
+
+  <Appenders>
+    <RollingFile name="RollingFile" fileName="atlas-cassandra-test.log">
```
Dalamar42: Ditto about logging to a file from tests

pom.xml
```diff
@@ -256,11 +256,39 @@
...
+        <version>2.3</version>
+      </dependency>
+
+      <dependency>
+        <groupId>org.atlasapi</groupId>
+        <artifactId>log4j2-logstash-jsonevent-layout</artifactId>
+        <version>1.0-SNAPSHOT</version>
+      </dependency>
+
       <dependency>
```
Dalamar42: Deer should already have a dependency to Kafka from `common-queue`. I'd avoid duplicating it because we are likely to get weird conflicts if the versions diverge

atlas-api/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration status="warn" name="Atlas-API" packages="*">
+
+  <Appenders>
```
Dalamar42: I don't see the Kafka appender here. Are you going to add it later?

Dalamar42: Never mind. Just reread your comments

Dalamar42:
FYI the log configuration are files are deployed by puppet rather than getting packaged in the war

atlas-api/src/main/resources/log4j2.xml
emilssolmanis: Should be trivial enough to merge XML via `xmllint` or a 10-line Py script if we can't.

atlas-api/src/test/resources/log4j2.xml
emilssolmanis: I'm actually pro this, keeps the surefire output cleaner and less problems with checking the exceptions later.

pom.xml
cognitive-i: TBH I want to pull all of the logging deps into a separate artefact, say common logging or something and hide that all there.  Doesn't smell right to put a dep to common-queue for pulling out kafka version.  I agree with the problem

cognitive-i:
Correct.  However Atlas code had 4 log4j properties files for randomised local testing.  I've just replaced them with rough log4j2 equivalents.  Happy to switch them to console since I don't believe they are used by puppet.

atlas-api/src/main/resources/log4j2.xml
Dalamar42: True, but the jetty script is standardised across all JVMs so we have to make sure it keeps supporting both

atlas-api/src/test/resources/log4j2.xml
Dalamar42: Yeah that could be a benefit though we can disable logging for mvn builds completely using environment variables to set logging level to OFF. Also there is a marginal benefit to being able to easily see logs in the stdout when running unit tests locally.

It's marginal though so I'm half and half on this

pom.xml
Dalamar42: Ah I see. Problem is I'm not sure Kafka's latest driver is backwards compatible so we might have some difficulty switching to it if it's not and the versions conflict.

emilssolmanis: Wait, are we still on 0.7? That's the only place where they went binary incompat, 0.7 -> 0.8.

Dalamar42: No we're on 0.8.1
We should be fine then

cognitive-i:
Review comments only

PR #298, opened 2016-02-26
Remove non actively published broadcasts when denormalising on ES series

Dalamar42:
- Change denormalisation logic to check if a broadcast is actively
published before denormalising it
- Change denormalisation logic to remove existing broadcasts from
parent series that have the same channel ID and start/end
transmission times as item broadcasts before merging them with the
item broadcasts. This is to effectively remove broadcasts belonging
to this item if they are no longer actively published
- Move logic to get possible parent series from ES and to reindex a
series on ES out of the EsContentTranslator and into
EsUnequivalentContentIndexer to improve separation of concerns and to
make testing easier
- Change parsing of ES broadcast channel ID to expect long since IDs
are longs
- Fix Failing test in PseudoEquivalentContentIndexIT


emilssolmanis:
:+1:

PR #299, opened 2016-03-01
Fix scariest and some scary FindBugs bugs

Dalamar42:
- SearchController: withCurrentBroadcastsOnly expects a primitive
boolean. Passing null with throw a RuntimeException. Change to
passing false instead if not specified
- User: manages method was checking applicationIds contains itself
rather than the provided applicationId. This code is currently unused
- Actor: Fix inconsistent equals/hashCode and change equals to follow
usual patterns rather than doing an instanceof to a specific type
that breaks symmetry of equals with superclass' equals
- KeyPhrase: String.format was expecting 3 parameters, but only two
are passed
- ChannelSchedule: Remove misleading final on a static field; Fix
inconsistent equals/hashCode and non-standard equals as before
- ScheduleRef: Fix inconsistent equals/hashCode and non-standard equals
as before
- EsContentTranslator: Fix possible NPE if existing series does not
have locations and returns null instead of the empty list

### DIFF
--- a/atlas-core/src/main/java/org/atlasapi/application/users/User.java
+++ b/atlas-core/src/main/java/org/atlasapi/application/users/User.java
@@ -105,7 +105,7 @@ public class User implements Identifiable {
     }
 
     public boolean manages(Id applicationId) {
-        return applicationIds.contains(applicationIds);
+        return applicationIds.contains(applicationId);
     }
 
     public boolean manages(Optional<Publisher> possibleSource) {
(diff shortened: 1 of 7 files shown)


atlas-core/src/main/java/org/atlasapi/application/users/User.java
```diff
@@ -105,7 +105,7 @@ public boolean manages(Application application) {
     }
 
     public boolean manages(Id applicationId) {
-        return applicationIds.contains(applicationIds);
```
emilssolmanis: didneyworl

emilssolmanis:
lgtm

PR #301, opened 2016-03-08
MBST-14517: Tighten RequestPath in * Controllers so return 404 

cognitive-i:
in case of broken path rather than 500 from ResponseWriterFactory.

Changed wildcard matcher to alpha extension matcher:

i.e.
`schedules.*` to `schedules\\.[a-z]+`
so match schedules.xml & schedules.json, but not schedules.hk45.json

### DIFF
--- a/atlas-api/src/main/java/org/atlasapi/query/v4/channel/ChannelController.java
+++ b/atlas-api/src/main/java/org/atlasapi/query/v4/channel/ChannelController.java
@@ -46,7 +46,7 @@ public class ChannelController {
         this.resultWriter = checkNotNull(resultWriter);
     }
 
-    @RequestMapping({ "", ".*", "/{id}.*", "/{id}" })
+    @RequestMapping({ "", "\\.[a-z]+", "/{id}\\.[a-z]+", "/{id}" })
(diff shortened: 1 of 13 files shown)


Dalamar42:
I'm not sure about this fix. We could have IDs that match [a-z] in which case this will still throw a 500.
Could we see which extensions we support and explicitly match them in the regex instead?

cognitive-i:
I disagree with the conclusion:
`
-    @RequestMapping({ "/{id}.*", "/{id}", ".*", "" })
+    @RequestMapping({ "/{id}\\.[a-z]+", "/{id}", "\\.[a-z]+", "" })
`

IDs needs a leading slash to match the RequestMapping.  I placed the `\\.` to break the reg-ex any char matcher.

cognitive-i:
Incidentally I didn't want to leak private implementation details of ResponseWriterFactory up further than I had to.  The other approach would be to push better exception throwing into ResponseWriterFactory.

Dalamar42:
Ah yes of course. You are excluding any extraneous "." in there. In that case LGTM

tfm:
:+1:

PR #303, opened 2016-03-11
Call start(), shutdown() on CachingChannelStore

Dalamar42:
- Not calling start(), shutdown() on PostConstruct, PreDestroy causes
 an NPE due to the underlying BackgroundComputingValue not
 instantiating properly


timothy22000:
LGTM. :+1:

emilssolmanis:
lgtm.

PR #307, opened 2016-04-12
Decomission event v1

Spartakovic:
-Removing old event model, store, transformers, serializers, protobufers, 
-Renaming eventv2 to event


atlas-api/src/test/java/org/atlasapi/output/annotation/EventAnnotationTest.java
```diff
@@ -1,95 +0,0 @@
...
-import org.mockito.runners.MockitoJUnitRunner;
-
-import static org.mockito.Matchers.any;
-import static org.mockito.Matchers.anyCollectionOf;
-import static org.mockito.Matchers.isA;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
-
-@RunWith(MockitoJUnitRunner.class)
-public class EventAnnotationTest {
```
Dalamar42: Why is this being deleted? This is an API annotation test

atlas-cassandra/src/main/resources/atlas_event_aliases.schema
```diff
@@ -1,10 +0,0 @@
-CREATE TABLE event_aliases (
```
Dalamar42: Do we have a new protobuf schema for event v2 aliases?

Dalamar42:
You need to rename the `/system` endpoints as well

Dalamar42:
LGTM

PR #311, opened 2016-04-26
Fix NPE in ChannelIntervalScheduleBootstrapTask

Dalamar42:
- ItemAndBroadcast null checks the passed in Item so there is no
  point continuing with this broadcast if we can't resolve a valid Item


emilssolmanis:
lgtm

PR #320, opened 2016-05-06
Fix user permission checking

Dalamar42:
- User A should have permission to edit user B if A == B OR if user A
  is an admin


Spartakovic:
LGTM

PR #321, opened 2016-05-09
Mbst 14272 add reviews ratings

cognitive-i:
* Extended model 
* Modified serialisation code
* Import from Owl Legacy transformer

TODO:
* make protocol buffers uses optional fields rather required

### DIFF
--- a/atlas-cassandra/src/main/java/org/atlasapi/entity/ReviewSerializer.java
+++ b/atlas-cassandra/src/main/java/org/atlasapi/entity/ReviewSerializer.java
@@ -3,28 +3,36 @@ package org.atlasapi.entity;
 import org.atlasapi.serialization.protobuf.CommonProtos;
 
 import java.util.Locale;
+import java.util.Optional;
 
 public class ReviewSerializer {
     public CommonProtos.Review serialize(Review review) {
-        CommonProtos.Review.Builder pbb = CommonProtos.Review.newBuilder();
+        CommonProtos.Review.Builder reviewBuilder = CommonProtos.Review.newBuilder();
 
         Locale locale = review.getLocale();
         if (null != locale) {
-            CommonProtos.LocaleString.Builder lppb = CommonProtos.LocaleString.newBuilder();
-            lppb.setValue(locale.getLanguage());
-            pbb.setLocale(lppb);
+            CommonProtos.LocaleString.Builder localeBuilder = CommonProtos.LocaleString.newBuilder();
+            localeBuilder.setValue(locale.getLanguage());
+            reviewBuilder.setLocale(localeBuilder);
         }
 
-        pbb.setReview(review.getReview());
-        return pbb.build();
+        reviewBuilder.setReview(review.getReview());
+        return reviewBuilder.build();
     }
 
-    public Review deserialize(CommonProtos.Review reviewPb) {
+    public Optional<Review> deserialize(CommonProtos.Review reviewBuffer) {
         Locale locale = null;
-        if (reviewPb.hasLocale()) {
-            locale = new Locale(reviewPb.getLocale().getValue());
+        if (reviewBuffer.hasLocale()) {
+            locale = new Locale(reviewBuffer.getLocale().getValue());
         }
 
-        return new Review(locale, reviewPb.getReview());
+        // all the fields of this protocol buffer are optional for future
+        // compatibility.  This incarnation requires .hasReview() to be true
+        // to match Review requirements
+        if (reviewBuffer.hasReview()) {
+            return Optional.of(new Review(locale, reviewBuffer.getReview()));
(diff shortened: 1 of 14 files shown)


atlas-cassandra/src/main/java/org/atlasapi/entity/ReviewSerializer.java
```diff
@@ -0,0 +1,30 @@
...
+            pbb.setLocale(lppb);
+        }
+
+        pbb.setReview(review.getReview());
+        return pbb.build();
+    }
+
+    public Review deserialize(CommonProtos.Review reviewPb) {
+        Locale locale = null;
+        if (reviewPb.hasLocale()) {
```
Dalamar42: We should avoid shorthand variable names like `reviewPb` or `pbb`.
You could do `reviewProtos` which in line with the lib naming and `reviewProtosBuilder` respectively.

Similarly for `lppb`.

atlas-cassandra/src/test/java/org/atlasapi/content/ContentSerializerTest.java
```diff
@@ -153,6 +158,15 @@ public void testDeSerializesFilm() {
         serializeAndCheck(film);
     }
 
+
+    @Test
+    public void testDeSerializesFilmWithReviewsAndRatings() {
```
Dalamar42: I think `Film` is unnecessary in the title. You are just using a Film to get around the fact Described is abstract, but it's not part of the test assertions.

atlas-core/src/main/java/org/atlasapi/entity/Rating.java
```diff
@@ -0,0 +1,52 @@
...
+
+        if (Float.compare(rating.value, value) != 0)
+            return false;
+        if (!type.equals(rating.type))
+            return false;
+        return publisher == rating.publisher;
+
+    }
+
+    @Override public int hashCode() {
```
Dalamar42: Use the Java 7 template for hashCode and equals instead as that's what we are using elsewhere

atlas-core/src/main/java/org/atlasapi/entity/Review.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.Locale;
+
+public class Review {
+
+    private final Locale locale;
+    private final String review;
+
+    public Review(@Nullable Locale locale, String review) {
+        this.locale = locale;
+        // note this is more strict than Owl (non-existent reviews should not be carried across)
```
Dalamar42: If you make Deer more strict than Owl you need to make sure the NPE thrown here will not prevent the containing Described or the other Review instances from being translated into Deer.

atlas-core/src/main/java/org/atlasapi/entity/Review.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        Review thatReview = (Review) that;
+
+        return Objects.equal(this.locale, thatReview.locale)
+                && Objects.equal(this.review, thatReview.review);
+    }
+
+    @Override
+    public int hashCode() {
```
Dalamar42: Similarly use the Java 7 template

atlas-legacy/src/test/java/org/atlasapi/system/legacy/LegacyContentTransformerTest.java
```diff
@@ -97,4 +102,60 @@ public void testIgnoreBroadcastsWithUnknownChannel() {
...
+        assertThat(transformedItem.getRatings().size(), is(2));
+
+        assertThat(transformedItem.getRatings().containsAll(Arrays.asList(
+                new org.atlasapi.entity.Rating("MOOSE", 1.0f, Publisher.BBC),
+                new org.atlasapi.entity.Rating("5STAR", 3.0f, Publisher.RADIO_TIMES)
+        )), is(true));
+    }
+
+
 }
```
Dalamar42: Newline at end of file

atlas-legacy/src/test/java/org/atlasapi/system/legacy/LegacyContentTransformerTest.java
```diff
@@ -97,4 +102,60 @@ public void testIgnoreBroadcastsWithUnknownChannel() {
 
         assertThat(transformed.getBroadcasts().size(), is(0));
     }
+
+    @Test
+    public void testReviews() {
```
Dalamar42: If you want to maintain Deer being more strict in handling reviews than Owl you should add a test here to ensure the translation doesn't fail for everything else

Dalamar42:
Some code style nitpicks, one genuine comment and the use of `optional` vs `required` in protobufs. Otherwise LGTM

PR #322, opened 2016-05-11
MBST-15350: LegacyContentTransformer: supports CrewMembers now

cognitive-i:


atlas-core/src/main/java/org/atlasapi/content/CrewMember.java
```diff
@@ -545,19 +546,22 @@ public Person toPerson() {
...
     @Override
-    public boolean equals(Object obj) {
-        if (obj instanceof CrewMember) {
-            CrewMember crew = (CrewMember) obj;
-            return this.getCanonicalUri().equals(crew.getCanonicalUri())
-                    && name.equals(crew.name)
-                    && role == crew.role;
-        }
-        return false;
+    public boolean equals(Object o) {
```
emilssolmanis: You didn't write this yourself, did you? It's a bit inconsistent, the generator should have an `Object.equals` dropdown option.

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyContentTransformer.java
```diff
@@ -583,6 +585,14 @@ private BrandRef transformToBrandRef(org.atlasapi.media.entity.ParentRef ref,
                         }
                 ).collect(Collectors.toList())
         );
+
+        c.setPeople(
+                input.people().stream()
+                        .map(legacyCrewMemberTransformer::apply)
+                        .filter(java.util.Objects::isNull)
```
emilssolmanis: My guess would be you want the exact opposite of this, this'll get you a list of `null`s.

atlas-core/src/main/java/org/atlasapi/content/CrewMember.java
Dalamar42: I think this is the Java 7 template from the IntelliJ generator which is what we use elsewhere

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyCrewMemberTransformer.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Objects;
+
+
+public class LegacyCrewMemberTransformer extends BaseLegacyResourceTransformer<org.atlasapi.media.entity.CrewMember, CrewMember>{
+
+
+    private void addCrewMember(org.atlasapi.media.entity.CrewMember legacy, CrewMember translated) {
+        translated.withName(legacy.name());
+
+        if (Objects.nonNull(legacy.role())) {
```
Dalamar42: This is a bit non-standard. We use `Objects.nonNull` in filters for Java 8 streams, but in this case I'd just do `legacy.role() != null`. It's more instantly recognisable

emilssolmanis: This is a bit unorthodox, generally `Objects.nonNull` is used for `.filter`, you can just do a simple null comparison here, i.e., `if (legacy.role() != null)`

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyCrewMemberTransformer.java
```diff
@@ -0,0 +1,61 @@
...
+            translated.withRole(CrewMember.Role.fromKey(legacy.role().key()));
+        }
+    }
+
+    private void addActor(org.atlasapi.media.entity.Actor legacy, Actor translated) {
+        translated.withCharacter(legacy.character());
```
Dalamar42: I take it, those `withX` method are mutating the object and not copying it. That name is a bit misleading

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyCrewMemberTransformer.java
```diff
@@ -0,0 +1,61 @@
...
+        addIdentified(legacy, translated);
+        addCrewMember(legacy, translated);
+        addActor(legacy, translated);
+
+        return translated;
+    }
+
+
+    public CrewMember translateLegacyCrewMember(org.atlasapi.media.entity.CrewMember legacy) {
+        if(! legacy.getClass().getName().equals("org.atlasapi.media.entity.CrewMember")) {
```
emilssolmanis: wat.

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyCrewMemberTransformer.java
```diff
@@ -0,0 +1,61 @@
...
+        addCrewMember(legacy, translated);
+
+        return translated;
+    }
+
+    @Override
+    public CrewMember apply(org.atlasapi.media.entity.CrewMember input) {
+        if (input instanceof org.atlasapi.media.entity.Actor) {
+            return translateLegacyActor((org.atlasapi.media.entity.Actor) input);
+        } else if(Objects.nonNull(input)) {
```
emilssolmanis: Same as above, just do a null comparison.

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyCrewMemberTransformer.java
Dalamar42: For what you want this is better: `legacy.getClass() != CrewMember.class`

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyCrewMemberTransformer.java
Dalamar42: Ditto `input != null`

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyContentTransformer.java
cognitive-i: I know exactly how that happened.  Originally I was using Optionals and filtering with present, then flipped to a lambda for the filter and then thought SURELY someone has done this before and then Guava got in the way

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyCrewMemberTransformer.java
cognitive-i: Afraid those with* APIs are nothing to do with me :)

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyCrewMemberTransformer.java
```diff
@@ -0,0 +1,71 @@
...
+        addIdentified(legacy, translated);
+        addCrewMember(legacy, translated);
+
+        return translated;
+    }
+
+    @Override
+    public CrewMember apply(org.atlasapi.media.entity.CrewMember input) {
+        // we carry IDs through from Owl to Deer, reject anything without one
+        if (null == input.getId()) {
```
emilssolmanis: Gonna guess that these are CPP habbits of the `3 == i` kind? It's a bit unusual for Java to have `null` first, not that it's incorrect or anything. :)

emilssolmanis:
lgtm, but that commit history looks a bit verbose, might wanna squash some. :shipit:

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyCrewMemberTransformer.java
Dalamar42: Yeah I know. It was a general comment :)

Dalamar42:
LGTM

PR #323, opened 2016-05-13
Refactor core annotation/application packages for code style/nullability

Dalamar42:
- Refactor code to conform to code style
- Switch from Guava functional style to Java8 streams
- Add `@Nullable` annotations and checkNotNull where appropriate.
  There was no attempt to fix null problems or to do other functional
  changes. This was intended only either enforce existing behaviour
  with respect to null or to surface the use of null so it can be
  tackled later with functional changes.
- Delete unused CacheBackedUserStore


Spartakovic:
lgtm

PR #324, opened 2016-05-13
Add persistent retry policy to all Kafka consumers

Dalamar42:
- This will save failed messages to mongo and retry them periodically
- Also refactor worker modules to have a consistent code style

### DIFF
--- a/atlas-processing/src/main/java/org/atlasapi/messaging/WorkersModule.java
+++ b/atlas-processing/src/main/java/org/atlasapi/messaging/WorkersModule.java
@@ -17,7 +17,6 @@ import org.atlasapi.system.bootstrap.workers.LegacyRetryingContentResolver;
 
 import com.metabroadcast.common.properties.Configurer;
 import com.metabroadcast.common.queue.Message;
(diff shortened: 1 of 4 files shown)


atlas-processing/src/main/java/org/atlasapi/messaging/WorkersModule.java
```diff
@@ -32,18 +31,25 @@
...
     private String consumerSystem = Configurer.get("messaging.system").get();
 
-    private String contentChanges = Configurer.get("messaging.destination.content.changes").get();
-    private String topicChanges = Configurer.get("messaging.destination.topics.changes").get();
-    private String scheduleChanges = Configurer.get("messaging.destination.schedule.changes").get();
-    private String contentEquivalenceGraphChanges = Configurer.get(
-            "messaging.destination.equivalence.content.graph.changes").get();
-    private String equivalentContentChanges = Configurer.get(
-            "messaging.destination.equivalent.content.changes").get();
+    private String contentChanges =
```
emilssolmanis: Presumably these should all be `final`?

emilssolmanis:
lgtm

PR #327, opened 2016-05-19
Filter unavailable `Location`s from API output.

emilssolmanis:
We've started handling revocations for availabilities in Nitro, which means we now use the `available`
flag on `Location`s. This makes the API output respect that flag.


Spartakovic:
lgtm

Dalamar42:
:shipit:

PR #328, opened 2016-05-19
Feature/channel filtering

Dalamar42:
The first commit is only refactoring. Most of it is IntelliJ refactors, but please have a look at the `ordering` method as that was written by hand. The second commit is the functional change.

### DIFF
--- a/atlas-api/src/main/java/org/atlasapi/query/v4/channel/ChannelQueryExecutor.java
+++ b/atlas-api/src/main/java/org/atlasapi/query/v4/channel/ChannelQueryExecutor.java
@@ -123,7 +123,7 @@ public class ChannelQueryExecutor implements QueryExecutor<Channel> {
                         input -> query
                                 .getContext()
                                 .getApplicationSources()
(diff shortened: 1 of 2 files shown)


atlas-api/src/main/java/org/atlasapi/query/v4/channel/ChannelQueryExecutor.java
```diff
@@ -133,46 +115,77 @@ public ChannelQueryExecutor(ChannelResolver resolver) {
...
+            );
+            break;
+        case TITLE_REVERSE:
+            ordering = MoreOrderings.transformingOrdering(
+                    input -> Strings.nullToEmpty(input.getTitle()),
+                    Ordering.<String>natural().reverse()
+            );
+            break;
+        default:
+            break;
```
emilssolmanis: This silently swallows invalid sort values. Generally, this should be an `IllegalArgumentException`, but depends on your case.

emilssolmanis:
:shipit:

PR #330, opened 2016-05-19
Mbst 15396 fix clip legacy importer

atish-metabroadcast:
Not ready for merge, but appreciate comments and thoughts about my mass commenting out and replumbing fest.  Tests are passing (but coverage is low)

### DIFF
--- a/atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyContentTransformer.java
+++ b/atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyContentTransformer.java
@@ -331,7 +331,7 @@ public class LegacyContentTransformer
 
     private Encoding transformEncoding(org.atlasapi.media.entity.Encoding input, Version version) {
         Encoding e = new Encoding();
-        setIdentifiedFields(e, input);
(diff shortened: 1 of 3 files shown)


atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyContentTransformer.java
```diff
@@ -597,6 +592,35 @@ private BrandRef transformToBrandRef(org.atlasapi.media.entity.ParentRef ref,
...
+        log.debug("Legacy '" + id + "': Clips out count=" + result.size());
+
+        return result;
+    }
+
+    private Optional<org.atlasapi.content.Clip> transformClip(Clip inputClip) {
+        try {
+            return Optional.of((org.atlasapi.content.Clip) apply(inputClip));
+        } catch(NullPointerException e) {
+            log.warn("Unable to process Clip of: "
```
emilssolmanis: Exceptions are formatted via placeholders, i.e., `log.warn("Unable to process {}", java.util.Objects.toString(...))`. Note that it'll also call `arg.toString()` automatically, so the conversion might not be necessary as well.

emilssolmanis:
:shipit:

PR #332, opened 2016-05-20
MBST-15466: Fix C* CrewMember asymmetric serialisation bug

atish-metabroadcast:
CrewMembers have a bunch of optional fields.
These are serialised without problems, but the deserialiser assumes presence of everything.

Owl only sparsely populates CrewMembers in most cases


Dalamar42:
:shipit:

PR #335, opened 2016-05-25
Updating Channel class to include the new fields.

MartinsIrbe:
- Adding new fields to Channel.
- Created ChannelType that is used by Channel to specify if the channel is channel or masterbrand.
- Wrote test for checking if the new fields are being set.
- Updating channel writer to write new channel fields.
- Adding tests to check if the channel fields are being written.


atlas-api/src/main/java/org/atlasapi/query/v4/channel/ChannelWriter.java
```diff
@@ -87,6 +87,12 @@ public void write(@Nonnull Channel entity, @Nonnull FieldWriter format,
         format.writeList(RELATED_LINKS_WRITER, entity.getRelatedLinks(), ctxt);
         format.writeField("start_date", entity.getStartDate());
         format.writeField("advertised_from", entity.getAdvertiseFrom());
+        format.writeField("short_description", entity.getShortDescription());
+        format.writeField("medium_description", entity.getMediumDescription());
+        format.writeField("long_description", entity.getLongDescription());
+        format.writeField("region", entity.getRegion());
+        format.writeList("target_regions", "target_regions", entity.getTargetRegions(), ctxt);
```
Dalamar42: `format.writeList("target_regions", "target_region", entity.getTargetRegions(), ctxt);`

Singular on the second parameter. That's the name for only one element

atlas-api/src/test/java/org/atlasapi/query/v4/channel/ChannelWriterTest.java
```diff
@@ -0,0 +1,262 @@
...
+
+        verify(fieldWriter).writeField(eq(title), eq(null));
+        verify(fieldWriter).writeField(eq(id), eq("pgnp"));
+        verify(fieldWriter).writeField(eq(uri), eq(null));
+        verify(fieldWriter).writeField(eq(mediaType), eq(null));
+        verify(fieldWriter).writeField(eq(highDefinition), eq(null));
+        verify(fieldWriter).writeField(eq(regional), eq(null));
+        verify(fieldWriter).writeField(eq(startDate), eq(null));
+        verify(fieldWriter).writeField(eq(advertisedFrom), eq(null));
+        verify(fieldWriter).writeField(eq(shortDescription), argumentCaptor.capture());
```
Dalamar42: What is the point of this `argumentCaptor`? You are already asserting the `shortDescription` is the expected one on the next line and you are not using this captor anyway?

atlas-api/src/test/java/org/atlasapi/query/v4/channel/ChannelWriterTest.java
```diff
@@ -0,0 +1,262 @@
...
+        this.channelWriter = new ChannelWriter(
+                channelGroupResolver,
+                listName,
+                fieldName,
+                channelGroupSummaryWriter
+        );
+    }
+
+    @Test
+    public void writingChannelShortDescription() throws IOException {
```
Dalamar42: Writing one test per field here is excessive and you are not even testing all the fields.

Write one test that sets all nullable values to null and make sure the fields written are null.
Then write another test that sets a valid value to all fields and make sure the expected values are written for all fields.

You don't need more than those two

atlas-core/src/main/java/org/atlasapi/channel/Channel.java
```diff
@@ -179,6 +197,36 @@ public DateTime getAdvertiseFrom() {
         return advertiseFrom;
     }
 
+    @FieldName("short_description")
+    public String getShortDescription() {
```
Dalamar42: All these getters may return null. Please mark them `@Nullable`

atlas-core/src/main/java/org/atlasapi/channel/Channel.java
```diff
@@ -66,7 +72,13 @@ public Channel(
             Set<ChannelRef> variations,
             @Nullable LocalDate startDate,
             @Nullable LocalDate endDate,
-            DateTime advertiseFrom
+            DateTime advertiseFrom,
```
Dalamar42: These fields except the Set can be null. Mark them `@Nullable`

atlas-core/src/main/java/org/atlasapi/channel/Channel.java
```diff
@@ -346,6 +399,36 @@ public Builder withAdvertiseFrom(DateTime dateTime) {
             return this;
         }
 
+        public Builder withShortDescription(String shortDescription) {
```
Dalamar42: All these with methods except the set can be null. Make the method parameters `@Nullable`

atlas-core/src/main/java/org/atlasapi/channel/ChannelType.java
```diff
@@ -0,0 +1,43 @@
...
+public enum ChannelType {
+
+    CHANNEL,
+    MASTERBRAND;
+
+    public String toKey() {
+        return this.name().toLowerCase();
+    }
+
+    private static final ImmutableSet<ChannelType> ALL = ImmutableSet.copyOf(values());
```
Dalamar42: Unless you want to use the `all()` method anywhere both `all()` and `ALL` are unnecessary. Inline them in the `KEY_MAP` generation

atlas-core/src/main/java/org/atlasapi/channel/ChannelType.java
```diff
@@ -0,0 +1,43 @@
...
+            .copyOf(Maps.transformValues(Maps.uniqueIndex(
+                    all(),
+                    new Function<ChannelType, String>() {
+                        @Override
+                        @Nullable
+                        public String apply(@Nullable ChannelType input) {
+                            return input.toKey();
+                        }
+                    }
+            ), Optional::ofNullable));
```
Dalamar42: Why are you doing this? `input.toKey()` can never return null. Simplify the map to `ImmutableMap<String, ChannelType>` and the `fromKey()` to `return Optional.fromNullable(KEY_MAP.get(key))`

atlas-core/src/test/java/org/atlasapi/channel/ChannelTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Before
+    public void setUp() throws Exception {
+        this.description = "Channel description.";
+        this.region = "region";
+        this.targetRegions = ImmutableSet.of("region 1", "region 2", "region 3");
+    }
+
+    @Test
+    public void shortDescriptionIsSetToChannel() {
```
Dalamar42: Again one test per field is excessive. Make one test to ensure you are coping with null values where you want to allow them and another one where you set and check all new fields

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyChannelTransformer.java
```diff
@@ -43,9 +46,25 @@ public Channel apply(org.atlasapi.media.channel.Channel input) {
...
     }
 
+    private ChannelType getChannelTypeFromInput(
+            org.atlasapi.media.channel.ChannelType inputChannelType
+    ) {
+        Optional<ChannelType> possibleChannelType = ChannelType.fromKey(inputChannelType.toKey());
+        if (possibleChannelType.isPresent()) {
+            return possibleChannelType.get();
+        }
+        return null;
```
Dalamar42: Don't return null here. If the channel type is not provided assume it's a channel. 
Also this will NPE if `inputChannelType` is null which can happen.

atlas-core/src/main/java/org/atlasapi/channel/Channel.java
Dalamar42: Except targetRegions and channelType

atlas-core/src/main/java/org/atlasapi/channel/Channel.java
```diff
@@ -45,6 +45,12 @@
...
     private final LocalDate endDate;
     private final DateTime advertiseFrom;
+    private final String shortDescription;
+    private final String mediumDescription;
+    private final String longDescription;
+    private final String region;
+    private final ImmutableSet<String> targetRegions;
+    private final ChannelType channelType;
 
     public Channel(
```
Dalamar42: if this constructor is not being used outside of this class please make it private

atlas-core/src/main/java/org/atlasapi/channel/Channel.java
```diff
@@ -88,6 +100,12 @@ public Channel(
         this.startDate = startDate;
         this.endDate = endDate;
         this.advertiseFrom = advertiseFrom;
+        this.shortDescription = shortDescription;
+        this.mediumDescription = mediumDescription;
+        this.longDescription = longDescription;
+        this.region = region;
+        this.targetRegions = ImmutableSet.copyOf(targetRegions);
+        this.channelType = channelType;
```
Dalamar42: Don't allow a null channel type. Everything should be providing this when building a channel and if they don't have it they should default to ChannelType.CHANNEL, but that's their responsibility

atlas-core/src/main/java/org/atlasapi/channel/Channel.java
Dalamar42: And channelType

atlas-core/src/main/java/org/atlasapi/channel/Channel.java
Dalamar42: And except channelType

atlas-core/src/main/java/org/atlasapi/channel/Channel.java
```diff
@@ -208,9 +256,14 @@ public String getKey() {
...
         private DateTime advertiseFrom;
-
         private LocalDate startDate;
         private LocalDate endDate;
+        private String shortDescription;
+        private String mediumDescription;
+        private String longDescription;
+        private String region;
+        private Set<String> targetRegions = Sets.newHashSet();
+        private ChannelType channelType;
```
Dalamar42: Default this to ChannelType.CHANNEL

atlas-legacy/src/test/java/org/atlasapi/system/legacy/LegacyChannelTransformerTest.java
```diff
@@ -35,6 +37,11 @@ public void testApply() throws Exception {
         org.atlasapi.media.entity.MediaType mediaType = org.atlasapi.media.entity.MediaType.VIDEO;
         Set<Publisher> availableFrom = ImmutableSet.of(Publisher.AMAZON_UK, Publisher.C4);
         DateTime advertiseFrom = DateTime.now();
+        String shortDescription = "Short description";
```
Dalamar42: Write another test here where all fields in the legacy channel that can be null, are null

atlas-api/src/test/java/org/atlasapi/query/v4/channel/ChannelWriterTest.java
MartinsIrbe: @Dalamar42 you are right, there is no point to have it.

atlas-core/src/main/java/org/atlasapi/channel/ChannelType.java
```diff
@@ -0,0 +1,40 @@
...
+                        @Override
+                        @Nullable
+                        public String apply(@Nullable ChannelType input) {
+                            return input.toKey();
+                        }
+                    }
+            ));
+
+    public static ChannelType fromKey(String key) {
+        ChannelType channelType = KEY_MAP.get(key);
```
Dalamar42: return Optional.ofNullable(KEY_MAP.get(key));

PR #337, opened 2016-05-25
MBST-15486: Add Reviews and Ratings to API

atish-metabroadcast:
note:
* Ratings have sources.

todo:
* Reviews do not have sources (in Owl Reviews inherit source from content).  Need to add same inherit functionality

### DIFF
--- a/atlas-api/src/main/java/org/atlasapi/output/writers/RatingsWriter.java
+++ b/atlas-api/src/main/java/org/atlasapi/output/writers/RatingsWriter.java
@@ -1,0 +1,44 @@
+package org.atlasapi.output.writers;
+
+import java.io.IOException;
+
+import javax.annotation.Nonnull;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import org.atlasapi.entity.Rating;
+import org.atlasapi.media.entity.Publisher;
+import org.atlasapi.output.EntityListWriter;
+import org.atlasapi.output.EntityWriter;
+import org.atlasapi.output.FieldWriter;
+import org.atlasapi.output.OutputContext;
+
+public class RatingsWriter implements EntityListWriter<Rating> {
+
+    private final EntityWriter<Publisher> sourceWriter;
+
+    public RatingsWriter(@Nonnull EntityWriter<Publisher> sourceWriter) {
+        this.sourceWriter = checkNotNull(sourceWriter);
+    }
+
+    @Nonnull
+    @Override
+    public String listName() {
+        return "ratings";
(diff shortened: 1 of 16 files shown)


atlas-api/src/main/java/org/atlasapi/output/writers/RatingsWriter.java
```diff
@@ -0,0 +1,44 @@
...
+import org.atlasapi.output.EntityListWriter;
+import org.atlasapi.output.EntityWriter;
+import org.atlasapi.output.FieldWriter;
+import org.atlasapi.output.OutputContext;
+
+public class RatingsWriter implements EntityListWriter<Rating> {
+
+    private final EntityWriter<Publisher> sourceWriter;
+
+    public RatingsWriter(@Nonnull EntityWriter<Publisher> sourceWriter) {
```
MartinsIrbe: Do we need `Nonnull` if we are then using `checkNotNull`?

atlas-cassandra/src/main/java/org/atlasapi/content/ContentDeserializationVisitor.java
```diff
@@ -89,7 +90,9 @@ public ContentDeserializationVisitor(ContentProtos.Content msg) {
 
     private <D extends Described> D visitDescribed(D described) {
         described = visitIdentified(described);
-        described.setPublisher(Sources.fromPossibleKey(msg.getSource()).get());
+
+        Optional<Publisher> possibleSource = Optional.ofNullable(Sources.fromPossibleKey(msg.getSource()).orNull());
+        described.setPublisher(possibleSource.get());  // NPE here is _desired behaviour_
```
MartinsIrbe: Why this is desired behaviour ?

MartinsIrbe:
2 comments otherwise LGTM 👍

atlas-api/src/main/java/org/atlasapi/output/writers/RatingsWriter.java
atish-metabroadcast: Nonnull is done by the IDE. ChecknotNull is done at runtime

atlas-cassandra/src/main/java/org/atlasapi/content/ContentDeserializationVisitor.java
atish-metabroadcast: This is the existing legacy publisher behaviour.  I only pull out the publisher to use latter on as part of the review transform.  See the previous line 92

PR #338, opened 2016-05-26
Feature/rt fields

Dalamar42:
This PR is to ensure that the Broadcast#newEpisode field gets ingested from Owl and appears in the API.

Please review commit by commit:

1. Cleanup as per code style -> Please review
2. Automatic re-order of code as per code style -> IntelliJ did this. A cursory look is enough
3. Code style; guava function removal and method ordering  -> A tiny bit I missed in 1. Please review
4. Import Broadcast#newEpisode field from Owl and add to API -> The actual feature
5. Clean up OutputContentMerger according to code style -> Mostly IntelliJ refactors. A cursory look please

### DIFF
--- a/atlas-api/src/main/java/org/atlasapi/output/writers/BroadcastWriter.java
+++ b/atlas-api/src/main/java/org/atlasapi/output/writers/BroadcastWriter.java
@@ -31,17 +31,25 @@ public final class BroadcastWriter implements EntityListWriter<Broadcast> {
 
     private static final String ELEMENT_NAME = "broadcast";
 
-    private final AliasWriter aliasWriter = new AliasWriter();
-    private final BroadcastIdAliasMapping aliasMapping = new BroadcastIdAliasMapping();
-    private final BlackoutRestrictionWriter blackoutRestrictionWriter = new BlackoutRestrictionWriter();
-    private final ChannelWriter channelWriter;
+    private final AliasWriter aliasWriter;
+    private final BroadcastIdAliasMapping aliasMapping;
+    private final BlackoutRestrictionWriter blackoutRestrictionWriter;
 
     private final String listName;
     private final NumberToShortStringCodec codec;
     private final ChannelResolver channelResolver;
+    private final ChannelWriter channelWriter;
 
-    public BroadcastWriter(String listName, NumberToShortStringCodec codec,
-            ChannelResolver channelResolver, ChannelGroupResolver channelGroupResolver) {
+    public BroadcastWriter(
+            String listName,
+            NumberToShortStringCodec codec,
+            ChannelResolver channelResolver,
+            ChannelGroupResolver channelGroupResolver
+    ) {
+        this.aliasWriter = new AliasWriter();
+        this.aliasMapping = new BroadcastIdAliasMapping();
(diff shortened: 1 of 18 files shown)


atlas-api/src/main/java/org/atlasapi/output/writers/BroadcastWriter.java
```diff
@@ -31,17 +31,25 @@
...
     private final ChannelResolver channelResolver;
+    private final ChannelWriter channelWriter;
+
+    public BroadcastWriter(
+            String listName,
+            NumberToShortStringCodec codec,
+            ChannelResolver channelResolver,
+            ChannelGroupResolver channelGroupResolver
+    ) {
+        this.aliasWriter = new AliasWriter();
```
emilssolmanis: Why pull these in here?

atlas-api/src/main/java/org/atlasapi/output/writers/BroadcastWriter.java
```diff
@@ -69,17 +86,17 @@ public void write(Broadcast entity, FieldWriter writer, OutputContext ctxt) thro
...
-        if (entity.getChannelId() != null) {
-            Optional<Channel> maybeChannel = Futures.get(channelResolver.resolveIds(ImmutableList.of(
-                    entity.getChannelId())), IOException.class)
-                    .getResources().first();
-            if (maybeChannel.isPresent()) {
-                channel = maybeChannel.get();
-            } else {
-                log.error("Unable to resolve channel {}", entity.getChannelId());
-            }
+        Optional<Channel> maybeChannel = Futures.get(channelResolver.resolveIds(ImmutableList.of(
```
emilssolmanis: This is still as fucked up, throw a ton of newlines at this and make it style-guide compliant (and therefore readable). Stuff like
```
Futures.get(
    channelResolver.resolveIds(
        ImmutableList.of(
```

atlas-api/src/main/java/org/atlasapi/output/writers/BroadcastWriter.java
Dalamar42: Consistency? That's how we do it elsewhere and it's nice to see how all dependencies are initialised in one place

atlas-api/src/main/java/org/atlasapi/output/writers/BroadcastWriter.java
Dalamar42: ack

atlas-api/src/main/java/org/atlasapi/output/writers/BroadcastWriter.java
emilssolmanis: Disagree, but fine. IMO it's more obvious that these are completely independent of anything if they're initialised in the way they were done before, but I don't care enough.

emilssolmanis:
I've no idea wth a "new episode" is, but technically lgtm. :shipit:

atlas-api/src/main/java/org/atlasapi/output/writers/BroadcastWriter.java
Dalamar42: The other reason I like this approach is that these things should be yank'able to allow tests to mock them. The fact we are not doing it here is another story.

PR #342, opened 2016-06-02
Fix bugs in no-auth admin endpoints

Dalamar42:
- No-auth endpoint to create source request succeeded if a null or
  invalid (not belonging to any user) email was provided. Now it will
  fail if it cannot resolve a valid user from the provided email or
  if an email is not provided at all
- When writing an application check if a revoked application is
  getting un-revoked. If so only allow it if the editing user is an
  admin
- Refactor several endpoints to ensure the old auth and the new
  no-auth endpoints call common code rather than being copy-paste


ljeyaratnam:
LGTM

PR #343, opened 2016-06-02
Added debug endpoint to get schedules from ScheduleStore

ljeyaratnam:
- moved existing code to debug equivalentSchdedules from ContentDebugController to ScheduleDebugController


Dalamar42:
LGTM

PR #344, opened 2016-06-03
Mbst 15351 topic genres updater

atish-metabroadcast:
Merging LegacyPersistentModule into AtlasPersistentModule.  I need this as PaTag mapper would create a horrific set of circular deps between those two modules.

LPM has a dependency on APM.  APM already contained a lot of LPM via copy and pasta to avoid circular dep.

Each commit tweaks the Modules pointing to LPM to point to APM.


emilssolmanis:
lgtm, but it's wiring. I.e., if it starts up, it's probably fine.

atlas-api/src/main/java/org/atlasapi/system/legacy/LegacyPersistenceModule.java
```diff
@@ -1,183 +0,0 @@
...
-public class LegacyPersistenceModule {
-
-    public static final String MONGO_COLLECTION_LOOKUP = "lookup";
-
-    @Autowired
-    private AtlasPersistenceModule persistence;
-
-    @Bean
-    @Qualifier("legacy")
-    public ContentResolver legacyContentResolver() {
```
JamieAP: Where'd this go?

Dalamar42:
LGTM

atlas-api/src/main/java/org/atlasapi/AtlasPersistenceModule.java
```diff
@@ -113,6 +114,8 @@
 public class AtlasPersistenceModule {
 
     private static final String MONGO_COLLECTION_LOOKUP = "lookup";
+    private static final String MONGO_TOPICS_COLLECTION = "topics";
```
JamieAP: MONGO_COLLECTION_LOOKUP and MONGO_TOPICS_COLLECTION :(

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyContentTransformer.java
```diff
@@ -620,7 +633,7 @@ private BrandRef transformToBrandRef(org.atlasapi.media.entity.ParentRef ref,
 
 
 
-    private Iterable<Tag> translateTopicRefs(List<TopicRef> topicRefs) {
+    private Iterable<Tag> translateTopicRefs(Collection<TopicRef> topicRefs) {
```
JamieAP: Prefer `Iterable` with `StreamSupport.stream(iterable.spliterator(), false)`

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyContentTopicMerger.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.stream.Collectors;
+
+import org.atlasapi.media.entity.TopicRef;
+
+import com.google.common.base.Equivalence;
+import com.google.common.collect.ImmutableSet;
+
+public class LegacyContentTopicMerger {
+
+    public static class TopicRefEquivalator extends Equivalence<TopicRef> {
```
Dalamar42: Code style please. Inner classes go last

atlas-legacy/src/main/java/org/atlasapi/system/legacy/PaTagMap.java
```diff
@@ -0,0 +1,247 @@
...
+
+    /** This method maps PA content genres with MetaBroadcast tags.
+     *
+     * This is done to get MetaBroadcast tags for a specific PA content. After that the
+     * PA content tags are added to the PA content topicRef field.
+     * @param genres - PA content genres that are used for mapping with MetaBroadcast tags.
+     * @return set of MetaBroadcast tags as TopicRef objects for the PA content.
+     */
+    public Set<TopicRef> mapGenresToTopicRefs(Set<String> genres) {
+        Set<String> tags = new HashSet<>(genres.size() + 1); // +1 to account for action film hack
```
JamieAP: `Sets` and `Maps` have useful static constructors fyi

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyContentTopicMerger.java
```diff
@@ -0,0 +1,45 @@
...
+        @Override
+        protected boolean doEquivalent(TopicRef a, TopicRef b) {
+            return ((a.getTopic() == b.getTopic())
```
Dalamar42: Does `getTopic()` return a string? If so equality should be checked with `equals` not `==`

JamieAP: `==` is checking pointer, you want .equals(), I think

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyContentTopicMerger.java
```diff
@@ -0,0 +1,45 @@
...
+                    && (a.getPublisher() == b.getPublisher()));
```
JamieAP: Same as above

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyContentTopicMerger.java
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        @Override
+        protected int doHash(TopicRef topicRef) {
+            return Objects.hash(topicRef.getTopic());
```
Dalamar42: Why are you not including the publisher in the hash? It doesn't violate the contract not to, but unless there is a specific reason not to include it, it would be more in line with the expectations of someone reading this code to have both `doEquivalent` and `doHash` use the same fields.

JamieAP:
Commented, OLGTM

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyContentTopicMerger.java
JamieAP: This should just return topicRef.hashCode(), imo. TopicRefs implementation should be responsible for making sure two TopicRefs pointing at the same Topic are equivalent.

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyContentTopicMerger.java
```diff
@@ -0,0 +1,45 @@
...
+        }
+    }
+
+    private TopicRefEquivalator equivalator = new TopicRefEquivalator();
```
Dalamar42: `private final`

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyContentTopicMerger.java
```diff
@@ -0,0 +1,45 @@
...
+
+    public List<TopicRef> mergeTags(Collection<TopicRef>... topicCollections) {
+
+        ImmutableSet.Builder<Equivalence.Wrapper<TopicRef>> setBuilder = new ImmutableSet.Builder<>();
+
+        setBuilder.addAll(
+                Arrays.asList(topicCollections).stream()
+                        .flatMap(topicCollection -> topicCollection.stream())
```
Dalamar42: You can use a method reference here

Dalamar42: Actually ignore me you can't

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyContentTopicMerger.java
```diff
@@ -0,0 +1,45 @@
...
+                        .map(equivalator::wrap)
+                        .collect(Collectors.toList()));
```
Dalamar42: `ImmutableCollectors.toSet()` directly

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyContentTopicMerger.java
```diff
@@ -0,0 +1,45 @@
...
+
+        @Override
+        protected int doHash(TopicRef topicRef) {
+            return Objects.hash(topicRef.getTopic());
+        }
+    }
+
+    private TopicRefEquivalator equivalator = new TopicRefEquivalator();
+
+    public List<TopicRef> mergeTags(Collection<TopicRef>... topicCollections) {
```
Dalamar42: Add a bit of Javadoc about the merge contract

atlas-api/src/main/java/org/atlasapi/AtlasPersistenceModule.java
atish-metabroadcast: Urgh, didn't notice that.  Sticky copy and pasta

PR #350, opened 2016-06-17
Adding priority_reasons annotation

MartinsIrbe:
This has been done so that we output priority reasons once the annotation has been provided.

### DIFF
--- a/atlas-api/src/main/java/org/atlasapi/output/annotation/ExtendedDescriptionAnnotation.java
+++ b/atlas-api/src/main/java/org/atlasapi/output/annotation/ExtendedDescriptionAnnotation.java
@@ -65,14 +65,10 @@ public class ExtendedDescriptionAnnotation extends OutputAnnotation<Content> {
         );
 
         if (ctxt.getActiveAnnotations().contains(Annotation.PRIORITY_REASONS)) {
-            if (desc.getPriority() == null) {
-                writer.writeObject(priorityReasonsWriter, null, ctxt);
+            if (desc.getPriority() != null && desc.getPriority().getReasons() != null) {
+                writer.writeObject(priorityReasonsWriter, desc.getPriority().getReasons(), ctxt);
             } else {
-                writer.writeObject(
-                        priorityReasonsWriter,
-                        desc.getPriority().getReasons() != null ? desc.getPriority().getReasons() : null,
-                        ctxt
-                );
+                writer.writeObject(priorityReasonsWriter, null, ctxt);
             }
         }
 
--- a/atlas-api/src/main/java/org/atlasapi/output/annotation/ExtendedDescriptionAnnotation.java
+++ b/atlas-api/src/main/java/org/atlasapi/output/annotation/ExtendedDescriptionAnnotation.java
@@ -6,6 +6,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 
+import org.atlasapi.annotation.Annotation;
 import org.atlasapi.content.Certificate;
 import org.atlasapi.content.Container;
 import org.atlasapi.content.Content;
@@ -16,6 +17,7 @@ import org.atlasapi.output.FieldWriter;
 import org.atlasapi.output.OutputContext;
 import org.atlasapi.output.writers.CertificateWriter;
 import org.atlasapi.output.writers.LanguageWriter;
+import org.atlasapi.output.writers.PriorityReasonsWriter;
 import org.atlasapi.output.writers.ReleaseDateWriter;
 import org.atlasapi.output.writers.RestrictionWriter;
 import org.atlasapi.output.writers.SubtitleWriter;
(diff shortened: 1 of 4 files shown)


atlas-api/src/main/java/org/atlasapi/output/annotation/ExtendedDescriptionAnnotation.java
```diff
@@ -59,6 +63,19 @@ public void write(Content desc, FieldWriter writer, OutputContext ctxt) throws I
...
                 desc.getPriority() != null ? desc.getPriority().getPriority() : null
         );
+
+        if (ctxt.getActiveAnnotations().contains(Annotation.PRIORITY_REASONS)) {
+            if (desc.getPriority() == null) {
+                writer.writeObject(priorityReasonsWriter, null, ctxt);
+            } else {
+                writer.writeObject(
+                        priorityReasonsWriter,
+                        desc.getPriority().getReasons() != null ? desc.getPriority().getReasons() : null,
```
emilssolmanis: This whole thing can be collapsed to
```
if (desc.getPriority() != null && desc.getPriority().getReasons() != null) {
    writer.writeObject(.., desc.getPriority().getReasons(), ...);
else {
    writer.writeObject(.., null, ...);
}
```
which makes it a bit clearer.

MartinsIrbe: 👍

atlas-api/src/main/java/org/atlasapi/output/writers/PriorityReasonsWriter.java
```diff
@@ -0,0 +1,37 @@
...
+        return "priority_reasons";
+    }
+
+    @Override
+    public void write(@Nonnull PriorityScoreReasons entity, @Nonnull FieldWriter writer,
+            @Nonnull OutputContext ctxt) throws IOException {
+        writer.writeList(
+                "positive",
+                "positive",
+                entity.getPositive() != null ? entity.getPositive() : null,
```
emilssolmanis: This is pointless, it's equivalent to just passing `entity.getPositive()`.

atlas-api/src/main/java/org/atlasapi/output/writers/PriorityReasonsWriter.java
```diff
@@ -0,0 +1,37 @@
...
+                ctxt
+        );
+
+        writer.writeList(
+                "negative",
+                "negative",
+                entity.getNegative() != null ? entity.getNegative() : null,
```
emilssolmanis: Same, this is equivalent to just passing `entity.getNegative()`.

atlas-core/src/main/java/org/atlasapi/annotation/Annotation.java
```diff
@@ -70,7 +70,8 @@
     SUPPRESS_EPISODE_NUMBERS,
     NON_MERGED,
     REVIEWS,
-    RATINGS;
+    RATINGS,
+    PRIORITY_REASONS;
```
emilssolmanis: Check the style guide on enums and fix this please, it's a 2-line diff here because the code doesn't adhere to the style guide.

emilssolmanis:
Minor cosmetics aside, this seems to make `priority_reasons` depend hierarchically on `extended_description`. Is that desired behaviour and something we've agreed upon?

MartinsIrbe:
@emilssolmanis yes as there is no point to display `priority_reasons` if `priority` score is a part of `extended_description`

emilssolmanis:
@MartinsIrbe Ok, but in that case, didn't we have some kind of hierarchy mechanism for exactly this kind of crap as opposed to manual `if``s inside the extended_description implementation?

emilssolmanis:
@MartinsIrbe Yup, check `QueryWebModule` for examples. Imo this is dirty and we should do this via the `implied` stuff we have, and `priority_reasons` should _imply_ `extended_description`. Would like @Dalamar42's opinion on this as well.

Dalamar42:
@emilssolmanis @MartinsIrbe I agree that if we want to establish a relationship between `extended_description` and `priority_reasons` then we should use the `imply` mechanism. However I don't agree that `priority_reasons` should imply `extended_description` here. That is not an obvious implication and `extended_description` implying `description` is already confusing for many people. For this to be intuitive you would have to do something like `extended_description_with_priority_reasons` which is ugly.

Unfortunately I can't think of a better way to do this. It's being done this way elsewhere as well.

The problem is the code was built to support:
- Annotations that write some data
- Annotations that imply other annotations
- Annotations that suppress other annotations

And now we're adding annotations that modify other annotations. We don't have a clean mechanism for the latter which is why this looks a bit nasty. Long term I think we should add support for such annotations at a higher level.

emilssolmanis:
@MartinsIrbe right, so Thomas says we do this in other places. Not that I like that, but lgtm once the other minor stuff is fixed.

MartinsIrbe:
@emilssolmanis 👍

PR #352, opened 2016-06-22
MBST-15351: TopicGenres: add tests, bug fix action films and tag empty genre items

atish-metabroadcast:
* tests showed typo in original PA tag mapper
* items with no genres are tagged with "unclassified" this allows us to eliminate them from genre bootstrap activities and perform genre improvement queries


Dalamar42:
LGTM although I don't see the logic that removes the unclassified tag once tags arrive. Will that be a separate PR?

atish-metabroadcast:
An additional PR isn't required.  The genres are pulled in wholesale and so the TopicRefs are overwritten every time there's update.  See

org/atlasapi/system/legacy/LegacyContentTransformer.java:569

Dalamar42:
Ack. Got it

PR #353, opened 2016-06-24
Simplify result reporting in ContentBootstrapListener

Dalamar42:
- The bootstrap process is fairly deeply nested. As a result the code
  that was recording success/failure messages for each step of the
  process ended up being very convoluted and hard to follow. Replace
  all that code with a graph that allows parts of the process to
  push/pop arbitrary paths in the graph and add success/failure
  messages under those paths.


Spartakovic:
lgtm

PR #354, opened 2016-06-27
Schedule: 'now' based queries could be racy between from/to

atish-metabroadcast:
Inject the same now into period parsers


Dalamar42:
LGTM

PR #355, opened 2016-06-30
Feature/content hashing

Dalamar42:


atlas-core/src/main/java/org/atlasapi/hashing/HashValueExtractor.java
```diff
@@ -0,0 +1,224 @@
...
+                .map(field -> getFieldObject(field, fieldContainingObject))
+                .filter(Objects::nonNull)
+                .map(this::getValueToHashFromObject)
+                .collect(Collectors.joining(FIELD_DELIMITER));
+    }
+
+    private ImmutableList<Field> getFields(Object object) {
+        ImmutableList.Builder<Field> fieldsBuilder = ImmutableList.builder();
+
+        for (Class<?> clazz = object.getClass(); clazz != null; clazz = clazz.getSuperclass()) {
```
MartinsIrbe: Maybe using Java 8 streams would be nicer :)

atlas-core/src/test/java/org/atlasapi/hashing/HashValueExtractorTest.java
```diff
@@ -0,0 +1,863 @@
...
+        assertThat(firstValue, is(secondValue));
+    }
+
+    @Test
+    public void getValueProducesExpectedResult() throws Exception {
+        AllSupportedFieldsTestClass testClass = new AllSupportedFieldsTestClass();
+
+        String value = hashValueExtractor.getValueToHash(testClass).get();
+
+        String expectedValue = ""
```
MartinsIrbe: Wouldn't it be better to read this string from a file?

atlas-core/src/main/java/org/atlasapi/hashing/HashValueExtractor.java
```diff
@@ -0,0 +1,224 @@
...
+
+    private String getValueToHashInternal(@Nullable Object object) {
+        if (object == null) {
+            return "";
+        }
+        return getValueToHashFromObject(object);
+    }
+
+    private String wrapValueWithTypeInformation(Object object, String valueToHash) {
+        return object.getClass().getCanonicalName()
```
emilssolmanis: `String.format` and making the format string a constant instead of each of the delimiters being a separate one might make this more readable.

atlas-core/src/main/java/org/atlasapi/hashing/HashValueExtractor.java
emilssolmanis: [Apache BeanUtils](http://commons.apache.org/proper/commons-beanutils/). 'nuff said.

atlas-core/src/main/java/org/atlasapi/hashing/HashValueExtractor.java
```diff
@@ -0,0 +1,224 @@
...
+    private String getValueToHashFromObject(Object object) {
+        Optional<String> extractedValue;
+
+        if (object instanceof Hashable) {
+            extractedValue = Optional.of(getValueToHashFromFields(object));
+        } else if (object.getClass().isPrimitive() ||
+                nativelySupportedTypes.contains(object.getClass())) {
+            extractedValue = Optional.of(object.toString());
+        } else if (object instanceof Enum) {
+            extractedValue = Optional.of(((Enum) object).name().toLowerCase());
```
emilssolmanis: This is vulnerable to different enum classes with equal value names. I.e.,
```
public enum Foo {
    ONE,
    TWO;
}

public enum Bar {
    ONE,
    NOPE_NOPE;
}

public class FooBarAbuse {
    private final Foo foo = Foo.ONE;
    private final Bar bar = Bar.ONE;
}
```
The field values will be indistinguishable, you need to namespace them.

atlas-core/src/main/java/org/atlasapi/hashing/HashValueExtractor.java
emilssolmanis: Actually nvm, I just found the abuse with resetting `private` fields that you're doing. Naughty....

emilssolmanis: Actually, if you're using fields, you should be using getters anyway, no? Let's discuss this when you're back.

atlas-core/src/main/java/org/atlasapi/hashing/HashValueExtractor.java
Dalamar42: All fields are serialised as key/value pairs where key is the fully qualified class name so they are effectively already namespace'd. This happens in `wrapValueWithTypeInformation`.

Also, it wouldn't matter because I am not aiming to make a bidirectional serialiser/deserialiser here. If both values where indistinguishable and then one or both changed then the hash would change which is sufficient here.

atlas-core/src/main/java/org/atlasapi/hashing/HashValueExtractor.java
Dalamar42: Fixed

atlas-core/src/test/java/org/atlasapi/hashing/HashValueExtractorTest.java
Dalamar42: Not in this case. There are no newlines in that string which would make it unreadable in a file

Dalamar42:
@emilssolmanis The existing timer already calculates the rate for writes. I am using a meter instead of another timer here because I don't care about the whole histogram of the measurements, just the rates so I can compare with the writes

emilssolmanis:
@Dalamar42 ok, so we have both data points then, cool. LGTM.

PR #361, opened 2016-07-14
Change partition key for update content messages to equiv set ID

Dalamar42:
- Workers downstream of the content store operate on the entire
  equivalence set for each updated content at once. Changing the
  partition key to the set ID ensures that all messages of the same
  set are always processed by the same worker.
- This is motivated by an issue where multiple updates on different
  content belonging to the same equiv set could result in many/all
  downstream workers processing the same equiv set at the same time
  and therefore competing for the same locks. This would then slow
  processing down and cause a backlog.


emilssolmanis:
lgtm

PR #362, opened 2016-07-15
Ban content update messages with given resource ID

Dalamar42:
- There is a bug with the partitioning logic which means updates on
  one particular content ID--which has started getting updated very
  frequently recently--ends up on multiple worker threads. These
  threads then compete for the same lock and as a result only one
  thread can proceed while the others are blocked. This in turns
  slows down the worker and causes a backlog. There is a fix for the
  partitioning logic being deployed, however for the sanity of
  support over the weekend the affected ID will get temporarily banned.


emilssolmanis:
lgtm

PR #364, opened 2016-07-18
Replace synchronised with locks

Dalamar42:
- Due to the use of synchronised we are effectively making two
  separate workers single threaded even though we have dozens of
  threads for each. This is affecting our ability to update
  equivalent schedules in a streaming fashion and forcing us to do
  scheduled bootstraps to ensure data consistency
- This commit replaces the use of synchronised with a group lock. The
  lock is the concatenation of the channelId, source and date for each
  schedule day that we are updating.


emilssolmanis:
lgtm

PR #366, opened 2016-08-02
Rr bootstrap fix

Spartakovic:
We stored adjacents as set of resourceRefs, however when we change type of content (item to episode, episode to item, etc), we store both refs which are causing exception when writing adjacency lists. We change it from set to map: Id to ResourceRefs to preserve uniqueness


atlas-core/src/main/java/org/atlasapi/equivalence/AbstractEquivalenceGraphStore.java
```diff
@@ -332,10 +333,15 @@ private EquivalenceGraph graphFor(ResourceRef subject,
...
-        if (subject.equals(adj.getRef())) {
+        if (subject.getId().equals(adj.getRef().getId())) {
             result = updateSubjectAdjacents(adj, assertedAdjacents, sources);
         } else if (sources.contains(adj.getSource())) {
-            if (assertedAdjacents.contains(adj.getRef())) {
+            Set<Id> ids = assertedAdjacents
+                    .stream()
+                    .map(ResourceRef::getId)
+                    .distinct()
+                    .collect(Collectors.toSet());
```
Dalamar42: Nitpick: You could do `MoreCollectors.toImmutableSet()` here

atlas-core/src/main/java/org/atlasapi/equivalence/EquivalenceGraph.java
```diff
@@ -47,19 +52,17 @@ public static final Adjacents valueOf(ResourceRef subject) {
...
                 Set<ResourceRef> afferent) {
             this.subject = checkNotNull(subject);
             this.created = checkNotNull(created);
             checkArgument(efferent.contains(subject));
             checkArgument(afferent.contains(subject));
-            this.efferent = ImmutableSet.copyOf(Identifiables.orderById()
-                    .immutableSortedCopy(efferent));
-            this.afferent = ImmutableSet.copyOf(Identifiables.orderById()
-                    .immutableSortedCopy(afferent));
+            this.efferent = efferent.stream().collect(MoreCollectors.toImmutableMap(ResourceRef::getId, Function.identity()));
```
Dalamar42: newlines after `.stream()` for readability

atlas-core/src/test/java/org/atlasapi/equivalence/AbstractEquivalenceGraphStoreTest.java
```diff
@@ -126,6 +130,79 @@ public void testAddingAnEquivalentResource() throws WriteException {
...
+    @Test
+    public void testMakingExistingEquivalenceToEquivalateToItemWithChangedType() throws WriteException {
+        Item subject = new Item(Id.valueOf(10), Publisher.AMAZON_UK);
+        subject.setThisOrChildLastUpdated(new DateTime(DateTimeZone.UTC));
+        makeEquivalent(paItem, subject);
+        Episode changedSubject = new Episode(Id.valueOf(10), Publisher.AMAZON_UK);
+        changedSubject.setThisOrChildLastUpdated(new DateTime(DateTimeZone.UTC));
+        makeEquivalent(paItem, changedSubject);
+
+        assertAfferentAdjacent(subject, paItem);
```
Dalamar42: Shouldn't this test be failing now? This looks like it asserts we should have a duplicate equivalences to both `subject` and `changedSubject` which is what this change is intended to fix.

Dalamar42:
You also need the serialisation test to ensure you can deserialise a JSON that was serialised with the old code.

PR #370, opened 2016-08-05
Schedule returns non merged broadcasts 

Spartakovic:
when provided with  non_merged annotation


emilssolmanis:
Where are we handling the other case? From what I see, this'll keep the current behaviour and do something weird if the annotation **is** specified.

Dalamar42:
As @emilssolmanis said where are we handling the case where the annotation is provided?

PR #374, opened 2016-08-09
Annotations access moved closer to persistence

Spartakovic:
-Added new deserialize method that accepts annotations that it can use after
-When deserializing check passed annotation when deserializing broadcasts, locations and sub_items
-Added new test to deserializing

### DIFF
--- a/atlas-cassandra/src/main/java/org/atlasapi/content/CassandraEquivalentContentStore.java
+++ b/atlas-cassandra/src/main/java/org/atlasapi/content/CassandraEquivalentContentStore.java
@@ -162,7 +162,7 @@ public class CassandraEquivalentContentStore extends AbstractEquivalentContentSt
 
         final SettableFuture<ResolvedEquivalents<Content>> result = SettableFuture.create();
 
-        resolveWithConsistency(result, ids, selectedSources, readConsistency);
+        resolveWithConsistency(result, ids, selectedSources, activeAnnotations, readConsistency);
 
         return result;
     }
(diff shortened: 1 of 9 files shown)


atlas-cassandra/src/main/java/org/atlasapi/content/CassandraEquivalentContentStore.java
```diff
@@ -225,7 +225,7 @@ public void onSuccess(Optional<ResolvedEquivalents<Content>> resolved) {
                             resolveWithConsistency(
                                     result,
                                     ids,
-                                    selectedSources,
+                                    selectedSources, activeAnnotations,
```
Dalamar42: newline

atlas-cassandra/src/main/java/org/atlasapi/content/CassandraEquivalentContentStore.java
```diff
@@ -287,6 +288,14 @@ public void onFailure(Throwable t) {
     }
 
     private Content deserialize(Row row) {
```
Dalamar42: Where is this used?

atlas-cassandra/src/main/java/org/atlasapi/content/ContentDeserializationVisitor.java
```diff
@@ -53,11 +68,20 @@
     private final ReviewSerializer reviewSerializer = new ReviewSerializer();
     private final RatingSerializer ratingSerializer = new RatingSerializer();
 
+    private final Set<Annotation> BROADCAST_ANNOTATIONS = ImmutableSet.of(BROADCASTS,
```
Dalamar42: private static final

Dalamar42: It's good your hoisted the broadcast annotations to a constant. For consistency it would be good to do the same for item refs and also locations (event though there is only one)

Dalamar42:
You should add tests for available and upcoming content and also add `@Nullable` annotations where appropriate in the model fields and getters/setters, but otherwise LGTM

PR #375, opened 2016-08-11
Incorporate code review comments

Dalamar42:
- Add class Javadoc
- Add meters to ContentHashGenerator to measure how often we failed
  to generate a hash
- Refactor HashValueExtractor to move extraction logic out to
  separate classes and methods to improve code readability

### DIFF
--- a/atlas-core/src/main/java/org/atlasapi/hashing/HashValueExtractor.java
+++ b/atlas-core/src/main/java/org/atlasapi/hashing/HashValueExtractor.java
@@ -4,31 +4,47 @@ import java.lang.reflect.Array;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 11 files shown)


atlas-core/src/main/java/org/atlasapi/hashing/HashValueExtractor.java
```diff
@@ -121,91 +131,37 @@ private String getValueToHashFromFields(Object fieldContainingObject) {
     }
 
     private String getValueToHashFromObject(Object object) {
```
Spartakovic: it is not the part of PR but, the name confuses me a little: get value to hash from object

Spartakovic:
one small comment, otherwise LGTM

atlas-core/src/main/java/org/atlasapi/hashing/HashValueExtractor.java
Dalamar42: yeah it's not great. I think I can do a better one here.

PR #379, opened 2016-08-12
Add logging for failing to resolve content during bootstrap

Dalamar42:
- We are getting quite a few of these failures. The logging is so we
  can capture the failing IDs


Spartakovic:
lgtm

PR #380, opened 2016-08-18
Mbst 16441/writing aliases

jamiesprax:
Gave ESLocation knowledge of aliases so that they can be recognised and searched with an ES query.

### DIFF
--- a/atlas-elasticsearch/src/test/java/org/atlasapi/content/EsLocationTest.java
+++ b/atlas-elasticsearch/src/test/java/org/atlasapi/content/EsLocationTest.java
@@ -1,0 +1,126 @@
+package org.atlasapi.content;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+import org.atlasapi.entity.Alias;
+
+import com.metabroadcast.common.stream.MoreCollectors;
+
+import org.joda.time.format.DateTimeFormat;
+import org.joda.time.format.DateTimeFormatter;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertThat;
+
+public class EsLocationTest {
+
+    private static final String AVAILABILITY_TIME = "availabilityTime";
+    private static final String AVAILABILITY_END_TIME = "availabilityEndTime";
+    private static final String ALIASES = "aliases";
+    private static final String NAMESPACE = "namespace";
+    private static final String VALUE = "value";
+
+    private static final String DATE_A = "16/08/2016 13:14:15";
+    private static final String DATE_B = "17/08/2016 14:15:16";
+
+    private DateTimeFormatter formatter;
+
+    private Map<String, Object> map;
+    private Map<String, Object> expectedMap = new HashMap<>();
+    private EsLocation expectedEsLocation;
+
+    private Alias aliasA, aliasB;
+    private List<Alias> aliases = new ArrayList<>();
+
+    @Before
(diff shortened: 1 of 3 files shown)


atlas-elasticsearch/src/test/java/org/atlasapi/content/EsLocationTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Test
+    public void aliasesTest() {
+
+        Map<String, Object> actualMap = new EsLocation().aliases(aliases).toMap();
+
+        Iterator<Map<String,Object>> actualAliasesIterator =
+                ((Iterable<Map<String,Object>>) actualMap.get(ALIASES)).iterator();
+
+        Map<String, Object> actualAliasMap;
```
Dalamar42: Avoid reusing variables when you can. You can do instead

```
Map<String, Object> actualAliasMapA = actualAliasesIterator.next();
// ...

Map<String, Object> actualAliasMapB = actualAliasesIterator.next();
// ...
```

atlas-elasticsearch/src/test/java/org/atlasapi/content/EsLocationTest.java
```diff
@@ -0,0 +1,108 @@
...
+        assertThat(actualAliasMap.get(VALUE), is(aliasA.getValue()));
+
+        actualAliasMap = actualAliasesIterator.next();
+        assertThat(actualAliasMap.get(NAMESPACE), is(aliasB.getNamespace()));
+        assertThat(actualAliasMap.get(VALUE), is(aliasB.getValue()));
+
+    }
+
+    @Test
+    public void fromMapTest() throws Exception {
```
Dalamar42: There is a bit of overlap between the two tests. The first one is covering the aliases and this one is covering both the aliases and the times. It's best to specialise tests so this one can be just doing date times.

atlas-elasticsearch/src/test/java/org/atlasapi/content/EsLocationTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+        EsLocation expectedESLocation = new EsLocation();
+        expectedESLocation.availabilityTime(formatter.parseDateTime(DATE_A).toDate());
+        expectedESLocation.availabilityEndTime(formatter.parseDateTime(DATE_B).toDate());
+        expectedESLocation.aliases(aliases);
+        expectedMap = expectedESLocation.toMap();
+    }
+
+    @Test
+    public void aliasesTest() {
```
Dalamar42: The new naming convention in junit tests is to avoid the word test in the name and also to say what the test does. In this case this could work: `addAliasesToEsLocation`

(`testAddingAliasesToEsLocation` or `addAliasesToEsLocationWorksCorrectly` is kinda implied, but in more complicated tests you would say something like `doingSomethingDoesSomethingElse` or somesuch)

Dalamar42:
Some small comments about naming and testing. OLGTM (otherwise looks good to me)

Dalamar42:
You are merging in the wrong branch though. You are merging in `MBST-16441-YV-Resolution-API`

Dalamar42:
You need to merge from your branch e.g. `MBST-16441/Writing-Aliases` to `MBST-16441/ResolvingLocationsByAliases` with the second branch being the one that will eventually get merged in master

atlas-elasticsearch/src/test/java/org/atlasapi/content/EsLocationTest.java
jamiesprax: The "fromMapTest()" tests the creation of a new ESLocation from a map, whereas the other tests that an already created ESLocation has been updated with the correct value.

PR #387, opened 2016-08-24
Feature/cql bootystrap

emilssolmanis:


atlas-processing/src/main/java/org/atlasapi/system/bootstrap/CqlContentBootstrapController.java
```diff
@@ -0,0 +1,314 @@
...
+    public void writeIdsToCqlStore(
+            HttpServletRequest request,
+            HttpServletResponse response
+    ) {
+        messageConsumerFactory.createConsumer(
+                new CqlWritingConsumer(),
+                JacksonMessageSerializer.forType(ResourceUpdatedMessage.class),
+                KAFKA_TOPIC_NAME,
+                KAFKA_CONSUMER_GROUP
+        ).withFailedMessagePersistence(mongo);
```
Dalamar42: the consumer is a Guava service. It needs to be started as well

atlas-processing/src/main/java/org/atlasapi/system/bootstrap/CqlContentBootstrapController.java
```diff
@@ -0,0 +1,314 @@
...
+        Optional<ContentListingProgress> storedProgress = progressStore.progressForTask(TASK_NAME);
+        ContentListingProgress progress = storedProgress.orElse(ContentListingProgress.START);
+
+        initSender();
+
+        listeningExecutorService.submit(() -> {
+            // Let's avoid sending brands and whatnot many times. Since each episode will reference
+            // its brand and series, this could be costly if we don't take care.
+            //
+            // For 50kk IDs, this should take
```
Dalamar42: I like `50kk` instead of `50M` :P

atlas-processing/src/main/java/org/atlasapi/system/bootstrap/CqlContentBootstrapController.java
```diff
@@ -0,0 +1,314 @@
...
+                sentAlready.set(toInt(brand));
+            }
+
+            if (series != null && ! sentAlready.get(toInt(series))) {
+                sendMessage(series, partition);
+                sentAlready.set(toInt(series));
+            }
+
+            sendMessage(episodeRef, partition);
+            sentAlready.set(toInt(episodeRef));
```
Dalamar42: Given you only check the Bloom filter for brands it might be good to not updated it with the IDs of series and episodes. Reduces the chance of a false positive.

Dalamar42:
OLGTM

atlas-processing/src/main/java/org/atlasapi/system/bootstrap/CqlContentBootstrapController.java
emilssolmanis: @Dalamar42 series too, basically anything that would get triggered via a `1-*` relationship, but good point.

PR #388, opened 2016-08-24
Feature/cql content consumer

emilssolmanis:


atlas-processing/src/main/java/org/atlasapi/system/bootstrap/workers/BootstrapWorkersModule.java
```diff
@@ -130,6 +131,30 @@ KafkaConsumer contentBootstrapWorker() {
...
+        );
+        return bootstrapQueueFactory()
+                .createConsumer(
+                        worker,
+                        new EntityUpdatedLegacyMessageSerializer(),
+                        contentChanges,
+                        "CqlContentBootstrap"
+                )
+                .withConsumerSystem(consumerSystem)
+                .withDefaultConsumers(contentChangesNumOfConsumers)
```
Dalamar42: Ideally those should be separate configuration values so you can tune this store's worker separately from the other one.

Dalamar42:
OLGTM

PR #389, opened 2016-08-24
Change Neo4j graph worker to get adjacents from LookupStore

Dalamar42:
- Not all equivalence graph changes in Owl will generate a Kafka
  message therefore if we only update according to what is in the
  message we will miss some adjacents
- Also resolving the adjacents as it arrives means we are more
  resilient to getting backlogged and less vulnerable to missing
  individual messages as we will always write the most up-to-date
  view of the equivalence graph with respect to the subject resourceRef


emilssolmanis:
lgtm

PR #390, opened 2016-08-24
Implement debug actions for CQL content

emilssolmanis:
Final nail in the CQL store feature.

The code itself is simple enough, my main concern is whether we want this as a separate controller or whether this actually belongs inside `ContentDebugController`. I'm somewhat more inclined to keep this separate because

- `ContentDebugController` already does too much, it has ID (de|en)coding and some other crap that has little to do with content
- `ContentDebugController` atm deals with mongo -> C*, adding a third store as a collaborator (plus maybe a dependency or two because of adding said store) seems to bloat it even further

I'd actually argue that I should write myself a ticket to go back and split up `ContentDebugController` at some later point, but I'm very welcome to be convinced otherwise, as the counterpoint I see here is duplication of concerns. Both of these deal with "debugging" (including fixing) content, just different stores.


Dalamar42:
LGTM

That's a good point actually about how cluttered that controller is. I just added some Neo4j stuff in there as well and it's looking very busy.

I think I'm in favour of splitting it up

PR #393, opened 2016-08-26
Bug/neo4j indices

Dalamar42:


emilssolmanis:
We can have a good ole fistfight about "indexes" vs. "indices", but olgtm. :shipit:

emilssolmanis:
lgtm

PR #396, opened 2016-09-05
Remove logic that generates Neo4j indices on boot

Dalamar42:
- This query is quite expensive for Neo4j even if the index and
  constraint already exist


emilssolmanis:
:shipit:

PR #397, opened 2016-09-05
Add metrics for each Neo4j writer to identify perf bottleneck

Dalamar42:


atlas-api/src/main/java/org/atlasapi/AtlasPersistenceModule.java
```diff
@@ -633,6 +633,6 @@ public Neo4jModule neo4jModule() {
 
     @Bean
     public Neo4jContentStore neo4jContentStore() {
-        return neo4jModule().neo4jContentStore();
+        return neo4jModule().neo4jContentStore(metricsModule.metrics());
```
emilssolmanis: Why not just wire the metrics module into neo4j module and just return a ready content store from there?

Dalamar42: The neo4j module is not an `@Configuration` class. I am continuing the pattern set by the ES module where the ES module itself is just a vanilla Java class and the atlas persistence module is the first Spring component in the wiring chain.

emilssolmanis: Meh. Fine.

emilssolmanis:
lgtm

PR #400, opened 2016-09-07
Persist locations with missing availability start/end times to ES

Dalamar42:
- The business logic behind locations is that a missing start dateTime
  implies a location has been available since forever while a missing
  end dateTime implies a location is available until forever. Before we
  were ignore locations with missing dateTimes which meant an
  actionable location call would mistakenly not return the items of
  those locations


emilssolmanis:
lgtm

PR #402, opened 2016-09-08
Change hierarchy writer to only write relationships in one direction

Dalamar42:
- We now only write in the direction episode -> series -> brand only.
  This is for performance reasons as we have content like flat brands
  with thousands of episodes that is very slow to update if all of
  these relationships have to be updated every time


atlas-neo4j/src/main/java/org/atlasapi/neo4j/service/writers/HierarchyWriter.java
```diff
@@ -21,6 +21,12 @@
 import static org.atlasapi.neo4j.service.model.Neo4jContent.HAS_BRAND_RELATIONSHIP;
 import static org.atlasapi.neo4j.service.model.Neo4jContent.HAS_SERIES_RELATIONSHIP;
 
+/**
+ * This class creates/updated the hierarchy relationships between brands, series and episodes.
```
emilssolmanis: update~~d~~**s**?

emilssolmanis:
lgtm

PR #403, opened 2016-09-14
Change channel_group endpoint to not show channel groups of channels

jamiesprax:
Currently this call is very expensive when used with the 'channel_groups_summary' annotation.

It returns all channels belonging to each returned channel group and also for each channel, every channel group that it belongs to.
This change removes the writing of the last part when the /channel_group/ endpoint is called.

The previous behaviour has been retained for the /channel/ endpoint.

### DIFF
--- a/atlas-api/src/main/java/org/atlasapi/query/QueryWebModule.java
+++ b/atlas-api/src/main/java/org/atlasapi/query/QueryWebModule.java
@@ -561,12 +561,24 @@ public class QueryWebModule {
         );
     }
 
-    private ChannelWriter channelWriter() {
-        return new ChannelWriter(
+    private ChannelWriter channelWriterGroupSummaryAnnotationSupported() {
+        return ChannelWriter.create(
                 channelGroupResolver,
                 "channels",
                 "channel",
-                new ChannelGroupSummaryWriter(idCodec())
+                new ChannelGroupSummaryWriter(idCodec()),
+                true
+
+        );
+    }
+
+    private ChannelWriter channelWriterGroupSummaryAnnotationNotSupported() {
+        return ChannelWriter.create(
+                channelGroupResolver,
+                "channels",
+                "channel",
+                new ChannelGroupSummaryWriter(idCodec()),
+                false
         );
     }
 
@@ -577,7 +589,7 @@ public class QueryWebModule {
                         CHANNELS,
                         new ChannelGroupChannelsAnnotation(
                                 new ChannelGroupChannelWriter(
-                                        channelWriter()
+                                        channelWriterGroupSummaryAnnotationNotSupported()
                                 ),
                                 channelResolver
                         ),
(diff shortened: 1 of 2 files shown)


atlas-api/src/main/java/org/atlasapi/query/QueryWebModule.java
```diff
@@ -561,12 +561,24 @@ ChannelGroupController channelGroupController() {
...
-    private ChannelWriter channelWriter() {
-        return new ChannelWriter(
+    private ChannelWriter channelWriterGroupSummaryAnnotationSupported() {
+        return ChannelWriter.create(
                 channelGroupResolver,
                 "channels",
                 "channel",
-                new ChannelGroupSummaryWriter(idCodec())
+                new ChannelGroupSummaryWriter(idCodec()),
+                true
```
emilssolmanis: Boolean arguments are bad style. They're completely unreadable, and they're quite the code smell generally because you might as well pass in a function instead of a flag.

One option is extracting an enum constant that implements a certain method, or (worse), given this already has 5 arguments, extract a builder and have a well named method on it that flips this flag.

atlas-api/src/main/java/org/atlasapi/query/v4/channel/ChannelWriter.java
```diff
@@ -51,16 +51,34 @@
     private final NumberToShortStringCodec idCode = SubstitutionTableNumberCodec.lowerCaseOnly();
     private final ChannelGroupSummaryWriter channelGroupSummaryWriter;
 
-    public ChannelWriter(
+    private final boolean channelGroupsSummaryAnnotationSupported;
+
+    private ChannelWriter(
```
emilssolmanis: As per above, 4 arguments was borderline. For 5, builder.

atlas-api/src/main/java/org/atlasapi/query/v4/channel/ChannelWriter.java
```diff
@@ -51,16 +51,34 @@
...
+            boolean channelGroupsSummaryAnnotationSupported
     ) {
         this.channelGroupResolver = checkNotNull(channelGroupResolver);
         this.listName = checkNotNull(listName);
         this.fieldName = checkNotNull(fieldName);
         this.channelGroupSummaryWriter = checkNotNull(channelGroupSummaryWriter);
+        this.channelGroupsSummaryAnnotationSupported = channelGroupsSummaryAnnotationSupported;
+    }
+
+    public static ChannelWriter create(ChannelGroupResolver channelGroupResolver,
```
emilssolmanis: Style, the 1st argument has to be on the next line.

atlas-api/src/main/java/org/atlasapi/query/v4/channel/ChannelWriter.java
```diff
@@ -51,16 +51,34 @@
...
+            String listName,
+            String fieldName,
+            ChannelGroupSummaryWriter channelGroupSummaryWriter,
+            boolean channelGroupSummaryAnnotationSupported
+    ) {
+        return new ChannelWriter(channelGroupResolver,
```
emilssolmanis: Same re style, args are either on one line or all on their own.

atlas-api/src/main/java/org/atlasapi/query/QueryWebModule.java
jamiesprax: Changed to use builder in commit c186c9d

PR #404, opened 2016-09-16
Fix for new constructor syntax

jamiesprax:
A class was not changed for new ChannelWriter constructor. It has now been fixed to use the new version.


MartinsIrbe:
LGTM

PR #405, opened 2016-09-16
Adding reporting to Columbus Telescope for content migration

MartinsIrbe:
This has been done so that we know more about content migration.

Updating the Jackson version to 2.8

Adding annotations to Id class so that we can get id as long and Long. Updated common dependency to the new version 2 as Timestamp millis method now also have Jackson annotations that we need in the deserialization logic.

Fixing failing tests as the deserialization logic for Joda intervals has changed and when deserializing interval, it's automatically converted to UTC.

### DIFF
--- a/atlas-core/src/main/java/org/atlasapi/entity/Id.java
+++ b/atlas-core/src/main/java/org/atlasapi/entity/Id.java
@@ -4,6 +4,10 @@ import java.math.BigInteger;
 
 import org.atlasapi.hashing.Hashable;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.google.common.base.Function;
 import com.google.common.primitives.Longs;
 
@@ -58,10 +62,13 @@ public final class Id implements Comparable<Id>, Hashable {
         this.longValue = longValue;
     }
 
+    @JsonCreator
     private Id(long longValue) {
         this.longValue = longValue;
     }
 
+    @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "@class")
+    @JsonSubTypes.Type(value = Long.class)
     public long longValue() {
         return longValue;
     }
--- a/atlas-processing/src/main/java/org/atlasapi/system/bootstrap/workers/BootstrapWorkersModule.java
+++ b/atlas-processing/src/main/java/org/atlasapi/system/bootstrap/workers/BootstrapWorkersModule.java
@@ -16,13 +16,20 @@ import org.atlasapi.messaging.v3.ScheduleUpdateMessage;
 import org.atlasapi.system.ProcessingHealthModule;
 import org.atlasapi.system.ProcessingMetricsModule;
 import org.atlasapi.system.bootstrap.ChannelIntervalScheduleBootstrapTaskFactory;
+import org.atlasapi.system.bootstrap.ColumbusTelescopeReporter;
 import org.atlasapi.system.bootstrap.EquivalenceWritingChannelIntervalScheduleBootstrapTaskFactory;
 import org.atlasapi.system.bootstrap.ScheduleBootstrapWithContentMigrationTaskFactory;
 
+import com.metabroadcast.columbus.telescope.client.TelescopeClientImpl;
 import com.metabroadcast.common.properties.Configurer;
 import com.metabroadcast.common.queue.kafka.KafkaConsumer;
 import com.metabroadcast.common.queue.kafka.KafkaMessageConsumerFactory;
 
+import com.codahale.metrics.MetricRegistry;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+import com.fasterxml.jackson.datatype.guava.GuavaModule;
+import com.fasterxml.jackson.datatype.joda.JodaModule;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
@@ -85,6 +92,10 @@ public class BootstrapWorkersModule {
     private final Boolean organisationBootstrapEnabled =
             Configurer.get("messaging.bootstrap.organisation.changes.enabled").toBoolean();
 
+    private final String columbusTelescopeHost =
+            Configurer.get("reporting.columbus-telescope.host").get();
+    private final String reportingEnvironment = Configurer.getPlatform();
+
     private final Set<Publisher> ignoredScheduleSources = Sets.difference(
             Publisher.all(),
             ImmutableSet.of(Publisher.PA, Publisher.BBC_NITRO, Publisher.BT_BLACKOUT)
@@ -110,12 +121,18 @@ public class BootstrapWorkersModule {
     @Bean
     @Lazy
     KafkaConsumer contentBootstrapWorker() {
+        MetricRegistry metricRegistry = metricsModule.metrics();
         ContentBootstrapWorker worker = ContentBootstrapWorker.create(
                 persistence.legacyContentResolver(),
                 persistence.contentStore(),
-                metricsModule.metrics().timer("ContentBootstrapWorker"),
-                metricsModule.metrics().meter("ContentBootstrapWorker.notWritten"),
-                metricsModule.metrics().meter("ContentBootstrapWorker.errorRate")
+                metricRegistry.timer("ContentBootstrapWorker"),
+                metricRegistry.meter("ContentBootstrapWorker.notWritten"),
+                metricRegistry.meter("ContentBootstrapWorker.errorRate"),
+                ColumbusTelescopeReporter.create(
+                        getTelescopeClient(),
+                        reportingEnvironment,
+                        getObjectMapper()
+                )
         );
         return bootstrapQueueFactory()
                 .createConsumer(
@@ -134,12 +151,18 @@ public class BootstrapWorkersModule {
     @Bean
     @Lazy
     KafkaConsumer cqlContentBootstrapWorker() {
+        MetricRegistry metricRegistry = metricsModule.metrics();
         ContentBootstrapWorker worker = ContentBootstrapWorker.create(
                 persistence.legacyContentResolver(),
                 persistence.cqlContentStore(),
-                metricsModule.metrics().timer("CqlContentBootstrapWorker"),
-                metricsModule.metrics().meter("CqlContentBootstrapWorker.notWritten"),
-                metricsModule.metrics().meter("CqlContentBootstrapWorker.errorRate")
+                metricRegistry.timer("CqlContentBootstrapWorker"),
+                metricRegistry.meter("CqlContentBootstrapWorker.notWritten"),
+                metricRegistry.meter("CqlContentBootstrapWorker.errorRate"),
+                ColumbusTelescopeReporter.create(
+                        getTelescopeClient(),
+                        reportingEnvironment,
+                        getObjectMapper()
+                )
         );
         return bootstrapQueueFactory()
                 .createConsumer(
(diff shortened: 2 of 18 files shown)


atlas-processing/src/test/java/org/atlasapi/system/debug/EventDebugControllerTest.java
```diff
@@ -41,21 +41,21 @@
...
-                    new JsonPrimitive(src.toString())
-    )
-            .create();;
+//    private NumberToShortStringCodec lowercaseDecoder;
+//    private Gson gson = new GsonBuilder().registerTypeAdapter(
+//            DateTime.class,
+//            (JsonSerializer<DateTime>) (src, typeOfSrc, context) ->
+//                    new JsonPrimitive(src.toString())
+//    )
+//            .create();;
```
MartinsIrbe: It's not being used, I will remove it. :)

atlas-processing/src/main/java/org/atlasapi/system/bootstrap/workers/BootstrapWorkersModule.java
```diff
@@ -85,6 +87,11 @@
     private final Boolean organisationBootstrapEnabled =
             Configurer.get("messaging.bootstrap.organisation.changes.enabled").toBoolean();
 
+    private final String columbusTelescopeHost =
+            Configurer.get("reporting.columbus-telescope.host").get();
+    private final String reportingEnvironment =
+            Configurer.get("reporting.columbus-telescope.environment").get();
```
emilssolmanis: This should come from `MBST_PLATFORM`, not as a separate config option that's bound to diverge.

atlas-processing/src/main/java/org/atlasapi/system/bootstrap/workers/BootstrapWorkersModule.java
```diff
@@ -115,7 +122,8 @@ KafkaConsumer contentBootstrapWorker() {
                 persistence.contentStore(),
                 metricsModule.metrics().timer("ContentBootstrapWorker"),
```
emilssolmanis: `metricsModule.metrics()` creates a new registry every time. Store the value.

atlas-processing/src/main/java/org/atlasapi/system/bootstrap/workers/ContentBootstrapWorker.java
```diff
@@ -88,9 +98,16 @@ private void process(Id contentId) throws org.atlasapi.entity.util.WriteExceptio
 
         if (result.written()) {
             log.debug("Bootstrapped content {}", result.toString());
+            columbusTelescopeReporter.reportSuccessfulMigration(
+                    content.getResources().first().get()
+            );
             time.stop();
         } else {
             log.debug("Content has not been written: {}", result.toString());
+            columbusTelescopeReporter.reportFailedMigration(
```
emilssolmanis: Is an unwritten content actually a failure? Wouldn't unchanged entities return false here as well?

atlas-processing/src/test/java/org/atlasapi/system/bootstrap/workers/ContentBootstrapWorkerTest.java
```diff
@@ -0,0 +1,210 @@
...
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ContentBootstrapWorkerTest {
+
+    private ContentResolver contentResolver = mock(ContentResolver.class);
```
emilssolmanis: We generally prefer the annotations to this.

atlas-processing/src/test/java/org/atlasapi/system/bootstrap/workers/ContentBootstrapWorkerTest.java
```diff
@@ -0,0 +1,210 @@
...
+        when(writer.writeContent(content)).thenReturn(successfulWriteResult);
+        when(metricsTimer.time()).thenReturn(timerContext);
+
+        bootstrapWorker.process(updatedMessage);
+
+        verify(telescopeClient, times(1)).createEvent(any(Event.class));
+        verify(telescopeClient).createEvent(columbusTelescopeEventCaptor.capture());
+
+        Event event = columbusTelescopeEventCaptor.getValue();
+        assertThat(event.getEntityState().get().getRaw().get(), is("{\"id\":{\"longValue\":[\"java.lang.Long\",1]},\""
```
emilssolmanis: This is just silly. Make the mapper injectable, mock this, and check for a sane value.

atlas-processing/src/test/java/org/atlasapi/system/bootstrap/workers/ContentBootstrapWorkerTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+                        @Override
+                        public ResourceType getResourceType() {
+                            return ResourceType.CONTENT;
+                        }
+                    }
+            );
+    }
+
+    private ListenableFuture<Resolved<Content>> createListenableFuture() {
```
emilssolmanis: wtf? mock? or `SettableFuture<?>`?

atlas-core/src/main/java/org/atlasapi/entity/Id.java
```diff
@@ -58,10 +62,13 @@ private Id(Long longValue) {
         this.longValue = longValue;
     }
 
+    @JsonCreator
     private Id(long longValue) {
         this.longValue = longValue;
     }
 
+    @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "@class")
+    @JsonSubTypes.Type(value = Long.class)
```
MartinsIrbe: @emilssolmanis I didn't said that I didn't add it there, what I said is that it works with `long` and `Long`.

atlas-core/src/main/java/org/atlasapi/equivalence/EquivalenceGraphUpdate.java
```diff
@@ -32,10 +32,10 @@
  */
 public class EquivalenceGraphUpdate {
 
-    private final EquivalenceGraph updated;
-    private final ImmutableSet<EquivalenceGraph> created;
-    private final ImmutableSet<Id> deleted;
-    private final EquivalenceAssertion assertion;
+    private EquivalenceGraph updated;
```
MartinsIrbe: @emilssolmanis if not making this mutable then Jackson can't deserialise this into a string and I need EquivalenceGraph as it's part of reporting (raw).

atlas-core/src/main/java/org/atlasapi/messaging/TimestampMixin.java
```diff
@@ -0,0 +1,15 @@
+package org.atlasapi.messaging;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+abstract class TimestampMixin {
```
MartinsIrbe: I don't need this, I will remove it :)

atlas-core/src/main/java/org/atlasapi/equivalence/EquivalenceGraphUpdate.java
emilssolmanis: But you have a `@JsonCreator` constructor. There's no reason this needs to be mutable.

MartinsIrbe: @emilssolmanis oh, yes you are right. Will fix this :)

atlas-processing/src/main/java/org/atlasapi/system/bootstrap/workers/ContentBootstrapWorker.java
MartinsIrbe: @emilssolmanis yes you are right. it's not a failure in that case.

atlas-processing/src/main/java/org/atlasapi/system/bootstrap/workers/BootstrapWorkersModule.java
MartinsIrbe: @emilssolmanis what do you mean by `MBST_PLATFORM`?

atlas-processing/src/test/java/org/atlasapi/system/bootstrap/workers/ContentBootstrapWorkerTest.java
MartinsIrbe: @emilssolmanis I can't mock this as `resourceRef#getTimestamp` is final, therefore I have to create actual `ResourceUpdatedMessage` object.

MartinsIrbe: I’m not quite sure what don’t you like about `ContentBootstrapWorkerTest#testReportingMigrationCorrectRawObjectIsSentAsPartOfTheEventReport` test? You want that I change the existing logic so that I can mock ObjectMapper however I don’t see point of doing that.

atlas-processing/src/main/java/org/atlasapi/system/bootstrap/workers/BootstrapWorkersModule.java
emilssolmanis: I mean the environment variable that we pass for every single project. There might even be some convenience methods around it in `Configurer`.

atlas-processing/src/test/java/org/atlasapi/system/bootstrap/workers/ContentBootstrapWorkerTest.java
emilssolmanis: @MartinsIrbe that's a giant hardcoded string there. It's brittle, it'll break any time anyone wants to change the serialisation even slightly, and fixing it will be a huge PITA. Mock the mapper to return some trivial thing and then compare against that.

MartinsIrbe: @emilssolmanis 👍

atlas-processing/src/test/java/org/atlasapi/system/bootstrap/workers/ContentBootstrapWorkerTest.java
```diff
@@ -0,0 +1,148 @@
...
+        this.bootstrapWorker = ContentBootstrapWorker.create(
+                contentResolver,
+                writer,
+                metricsTimer,
+                contentNotWrittenMeter,
+                failureMeter,
+                columbusTelescopeReporter
+        );
+        this.content = new org.atlasapi.content.Item("panda.com", "panda", Publisher.PA);
+        this.content.setId(Id.valueOf(1l));
```
emilssolmanis: Uppercase `L` for numbers. This will at the very least fail Sonar.

atlas-processing/src/test/java/org/atlasapi/system/bootstrap/workers/ContentBootstrapWorkerTest.java
```diff
@@ -0,0 +1,148 @@
...
+    public void testReportingMigrationCorrectRawObjectIsSentAsPartOfTheEventReport()
+            throws RecoverableException, WriteException, JsonProcessingException {
+        this.content.setLastFetched(DateTime.parse("2016-08-28T00:00:00Z"));
+        ResourceUpdatedMessage updatedMessage = createUpdatedMessage();
+
+        when(resourceRef.getId()).thenReturn(Id.valueOf(1));
+        when(contentResolver.resolveIds(any(Collection.class))).thenReturn(Futures.immediateFuture(Resolved.valueOf(ImmutableList.of(content))));
+        when(writer.writeContent(content)).thenReturn(successfulWriteResult);
+        when(metricsTimer.time()).thenReturn(timerContext);
+        when(objectMapper.writeValueAsString(any(Content.class))).thenReturn("{ \"test\":\"test\"}");
```
emilssolmanis: Magic string that's copy-pasted. Extract variable.

atlas-processing/src/test/java/org/atlasapi/system/bootstrap/workers/ContentBootstrapWorkerTest.java
```diff
@@ -0,0 +1,148 @@
...
+    @Mock private ContentResolver contentResolver;
+    @Mock private ContentWriter writer;
+    @Mock private Timer metricsTimer;
+    @Mock private Meter contentNotWrittenMeter;
+    @Mock private Meter failureMeter;
+    @Mock private TelescopeClientImpl telescopeClient;
+    @Mock private ResourceRef resourceRef;
+    @Mock private Timer.Context timerContext;
+    @Mock private ResourceUpdatedMessage resourceUpdatedMessage;
+    @Mock private ListenableFuture<Resolved<Content>> listenableFuture;
```
emilssolmanis: This seems to be unused.

PR #406, opened 2016-09-19
Improve Neo4jContentStore metrics

Dalamar42:
- Add metrics to every part of the Neo4jContentStore write pipelines
  to identify the bottleneck
- Move metric definition in the Neo4jContentStore constructor to
  simplify its builder
- Remove the metrics from the individual writers since they were not
  helpful in finding any bottleneck (individually they all appear to
  be fast)

### DIFF
--- a/atlas-neo4j/src/main/java/org/atlasapi/neo4j/service/Neo4jContentStore.java
+++ b/atlas-neo4j/src/main/java/org/atlasapi/neo4j/service/Neo4jContentStore.java
@@ -24,7 +24,8 @@ import org.atlasapi.neo4j.service.writers.EquivalenceWriter;
 import org.atlasapi.neo4j.service.writers.HierarchyWriter;
 import org.atlasapi.neo4j.service.writers.LocationWriter;
 
-import com.codahale.metrics.Histogram;
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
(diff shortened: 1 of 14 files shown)


atlas-neo4j/src/main/java/org/atlasapi/neo4j/service/Neo4jContentStore.java
```diff
@@ -198,10 +321,28 @@ public Void visit(Clip clip) {
...
+                () -> locationWriter.write(item, transaction),
+                writeContentLocationTimer
+        );
+        runWithTiming(
+                () -> broadcastWriter.write(item, transaction),
+                writeContentBroadcastTimer
+        );
+    }
+
+    private void runWithTiming(Runnable writer, Timer timer) {
```
emilssolmanis: This argument being a `Runnable` is a bit confusing. Just make it a `Supplier<Void>` (or similar, if you can think of a better option), especially since per its docs

> There is no requirement that a new or distinct result be returned each time the supplier is invoked.

Dalamar42: @emilssolmanis My understanding was that the reason they didn't add a function that accepts nothing and returns nothing in `java.util.function` was because `Runable` was intended to serve that role. Is that wrong?

Thing is if I change it to `Supplier<Void>` I will have to add a `return null` to every single lambda which is a bit ugly.

PR #407, opened 2016-09-20
Optimise Neo4j writes to reduce number of updates needed

Dalamar42:
- Change EquivalenceWriter writes to create resourceRef nodes where
  needed rather than executing a separate query to write them before
  writing equivalence assertions
- Change HierarchyWriter writes to create contentRef nodes where
  needed rather than delegating to ContentWriter to create them
- Use ON CREATE SET clauses in ContentWriter to avoid repeatedly
  updating data that cannot change
- Remove unnecessary metrics from Neo4jContentStore. The Neo4j driver
  does most of the work when the transaction is closed so that is where
  most of the execution time is concentrated. The metrics of the
  other execution steps are showing unrealistic low values that are
  not useful
- Remove methods to write contentRef and resourceRef from
  ContentWriter as they are no longer needed


emilssolmanis:
Kinda lgtm I guess.

PR #408, opened 2016-09-22
Add metrics to workers

Dalamar42:
- Add metrics to mark reception of message to Neo4j workers
- Enable Kafka consumer metrics for all workers


Dalamar42:
I have also introduced the new naming scheme for metrics. If you think it's good I will add it to the code style.

PR #410, opened 2016-09-28
Standardise metric naming

Dalamar42:
- Standardise metric naming across Kafka workers and in
  ContentHashGenerator for consistency and to make it easier to
  create standardised views for them in Grafana

### DIFF
--- a/atlas-processing/src/main/java/org/atlasapi/messaging/ContentEquivalenceUpdatingWorker.java
+++ b/atlas-processing/src/main/java/org/atlasapi/messaging/ContentEquivalenceUpdatingWorker.java
@@ -12,7 +12,7 @@ import com.metabroadcast.common.stream.MoreCollectors;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
-import com.google.api.client.repackaged.com.google.common.base.Throwables;
+import com.google.common.base.Throwables;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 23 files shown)


atlas-processing/src/main/java/org/atlasapi/system/bootstrap/workers/BootstrapWorkersModule.java
```diff
@@ -277,17 +306,21 @@ KafkaConsumer topicReadWriter() {
...
+                metricsModule.metrics()
         );
+
         return bootstrapQueueFactory()
                 .createConsumer(
                         worker,
                         new EntityUpdatedLegacyMessageSerializer(),
                         eventChanges,
-                        "SeparatingEventBootstrap"
+                        workerName
```
AdamDM: duplicate code? Extract the return block and pass what changes in

atlas-processing/src/main/java/org/atlasapi/messaging/ContentEquivalenceUpdatingWorker.java
```diff
@@ -15,6 +12,7 @@
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
+import com.google.api.client.repackaged.com.google.common.base.Throwables;
```
emilssolmanis: You're wrong.

Dalamar42: Err, yes... Fixing

atlas-processing/src/main/java/org/atlasapi/system/bootstrap/workers/BootstrapWorkersModule.java
Dalamar42: https://programmers.stackexchange.com/questions/300043/illusory-code-duplication

We already have the builder to standardise creating consumers. Adding a method to wrap that builder would be overengineering a bit in this case.

PR #417, opened 2016-10-07
Mbst 16562/refactor violations in deer architecture

jamiesprax:
ChannelGroup and Channel annotations were using resolvers inside annotations to retrieve the data they need. This should have been done in the executor as it violates the Deer architecture. The annotations that previously used resolvers have had their logic refactored to the executor where the annotation is read and appropriate content resolved. 

Tests have been written to check single & multi query execute queries.

### DIFF
--- a/atlas-api/src/main/java/org/atlasapi/output/annotation/ChannelGroupChannelsAnnotation.java
+++ b/atlas-api/src/main/java/org/atlasapi/output/annotation/ChannelGroupChannelsAnnotation.java
@@ -53,7 +53,7 @@ public class ChannelGroupChannelsAnnotation extends OutputAnnotation<ResolvedCha
                 .getParameter(Attributes.CHANNEL_GROUP_CHANNEL_GENRES.externalName());
 
         if (!Strings.isNullOrEmpty(genre)) {
-            final ImmutableSet<String> genres = ImmutableSet.copyOf(Splitter.on(',')
+            ImmutableSet<String> genres = ImmutableSet.copyOf(Splitter.on(',')
                     .split(genre));
             channels = Iterables.filter(channels,
                     input -> !Sets.intersection(input.getGenres(), genres).isEmpty()
--- a/atlas-api/src/main/java/org/atlasapi/output/annotation/ChannelGroupChannelsAnnotation.java
+++ b/atlas-api/src/main/java/org/atlasapi/output/annotation/ChannelGroupChannelsAnnotation.java
@@ -1,32 +1,25 @@
 package org.atlasapi.output.annotation;
 
 import java.io.IOException;
-import java.util.List;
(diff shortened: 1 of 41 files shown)


atlas-api/src/main/java/org/atlasapi/output/annotation/ChannelGroupChannelsAnnotation.java
```diff
@@ -1,94 +1,77 @@
...
-            builder.put(channelGroupMembership.getChannel().getId(), channelGroupMembership);
+
+        Optional<Iterable<ResolvedChannel>> resolvedChannels = entity.getChannels();
+
+        if(!resolvedChannels.isPresent()) {
+            throw new MissingResolvedDataException(channelWriter.listName());
         }
 
-        ImmutableMultimap<Id, ChannelGroupMembership> channelGroupMemberships = builder.build();
+        Iterable<Channel> channels = StreamSupport.stream(resolvedChannels.get().spliterator(), false)
```
jakekeeys: any reason why this can't be immutable

atlas-api/src/main/java/org/atlasapi/output/annotation/ChannelVariationAnnotation.java
```diff
@@ -1,61 +1,47 @@
...
-                    public Iterable<Channel> apply(@Nullable Resolved<Channel> input) {
-                        return input.getResources();
-                    }
-                }
-        ), 1, TimeUnit.MINUTES, IOException.class);
-        format.writeList(channelWriter, channels, ctxt);
+    public void write(ResolvedChannel entity, FieldWriter format, OutputContext ctxt) throws IOException {
+
+        Optional<Iterable<Channel>> channelVariations = entity.getChannelVariations();
+        if (!channelVariations.isPresent()) {
```
jakekeeys: Consistency: You handle exception in the body of the "else" elsewhere

atlas-api/src/main/java/org/atlasapi/output/writers/BroadcastWriter.java
```diff
@@ -99,9 +97,14 @@ public void write(Broadcast entity, FieldWriter writer, OutputContext ctxt) thro
         if (channel == null) {
             log.error("Unable to resolve channel {}", entity.getChannelId());
 
+        } else {
+            // Little hack until Broadcasts have their own composite objects to deal with resolution of
```
jakekeeys: Create TODO for this?

atlas-api/src/main/java/org/atlasapi/output/annotation/ChannelGroupChannelsAnnotation.java
jamiesprax: I have it mutable because its filtered on a few lines down

PR #420, opened 2016-10-11
Change how stale broadcast deletions are handled

Dalamar42:
- Only delete stale broadcasts from the intersection of the days in
  the broadcast interval and the schedule update interval. This way
  we will only delete a stale broadcast from a particular day if that
  stale broadcast's interval overlaps with that day and and schedule
  update we are currently processing also overlaps with that day.
  This protects from over-eagerly deleting a broadcast from a day
  that the update is not asserting on


atlas-testlib/src/main/java/org/atlasapi/schedule/EquivalentScheduleStoreTester.java
```diff
@@ -1578,7 +1590,9 @@ public void testDoesnRemoveStaleBroadcastsFromOutsideInterval() throws Exception
         EquivalentSchedule resolved = get(
                 getSubjectGenerator().getEquivalentScheduleStore().resolveSchedules(
                         ImmutableList.of(channel),
-                        new Interval(item1Broadcast1Start, item2BroadcastEnd),
+                        new Interval(new DateTime(2015, 10, 25, 19, 30, 0, 0, DateTimeZones.UTC),
```
emilssolmanis: And another newline.

PR #426, opened 2016-10-12
Bt annotation fix

jamiesprax:
This stop all ChannelGroups from being returned that do not have a whitelisted ID or Alias namespace. The whitelist was created from the docs page displaying the special BT channel group ID's and namespaces.


atlas-api/src/main/java/org/atlasapi/query/v4/channel/ChannelQueryExecutor.java
```diff
@@ -55,6 +58,39 @@
     private final ChannelResolver channelResolver;
     private final ChannelGroupResolver channelGroupResolver;
 
+    private final SubstitutionTableNumberCodec codec = SubstitutionTableNumberCodec.lowerCaseOnly();
```
Dalamar42: Why is this whitelist in the ChannelExecutor? The behaviour of the channel endpoint shouldn't change. This is only for the channel group endpoint

jamiesprax: I have removed the whitelist from the channel executor

PR #427, opened 2016-10-12
Hardcoded list of ID's to filter by before resolution

jamiesprax:
The desired effect was there which was removing channel groups
that were not needed however there was no performance gain as
all the channel groups still had to be resolved so we could
check their aliases. A list from the db has been hardcoded so now
we can filter before the ID's are passed to the resolver.

### DIFF
--- a/atlas-api/src/main/java/org/atlasapi/query/v4/channelgroup/ChannelGroupQueryExecutor.java
+++ b/atlas-api/src/main/java/org/atlasapi/query/v4/channelgroup/ChannelGroupQueryExecutor.java
@@ -21,7 +21,6 @@ import org.atlasapi.channel.ResolvedChannel;
 import org.atlasapi.channel.ResolvedChannelGroup;
 import org.atlasapi.criteria.AttributeQuery;
 import org.atlasapi.criteria.attribute.Attributes;
-import org.atlasapi.entity.Alias;
 import org.atlasapi.entity.Id;
 import org.atlasapi.entity.util.Resolved;
 import org.atlasapi.output.NotFoundException;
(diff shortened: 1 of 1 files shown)


atlas-api/src/main/java/org/atlasapi/query/v4/channelgroup/ChannelGroupQueryExecutor.java
```diff
@@ -82,10 +81,22 @@
...
-            "bt:subscription-code",
-            "bt:tug"
-    );
+//    Slow alias lookup fields, kept just in case new faster resolution doesn't work properly
+//    private final List<String> whitelistedNamespaces = Lists.newArrayList(
+//            "bt:subscription-code",
+//            "bt:tug"
+//    );
+
+    private final List<Long> btSubscriptionChannelGroupIds = Lists.newArrayList(103857L,104240L,104242L,104239L,104262L,104238L,104265L,104302L,104275L,104282L,104283L,104241L,104298L,104266L,104301L,104284L,104269L,104277L,104290L,104292L,104293L,104294L,104295L,104296L,104297L,104300L,104303L,104306L,104309L,104263L,104299L,104274L,104304L,104311L,104313L,104285L,104291L,104310L,104264L,104268L,104267L,104276L,104307L,104312L,104305L,104308L,104336L,104337L,104338L,104339L,104340L,104341L,104342L,104344L,104345L,104346L,104347L,104348L,104350L,104351L,104352L,104353L);
```
Dalamar42: Line length

atlas-api/src/main/java/org/atlasapi/query/v4/channelgroup/ChannelGroupQueryExecutor.java
```diff
@@ -82,10 +81,22 @@
...
+    private final List<Long> btTargetUserGroupChannelGroupIds = Lists.newArrayList(103860L,104243L,104343L,104349L);
+
+    private final ImmutableList<Long> whitelistedAliasLongs = ImmutableList.<Long>builder().addAll(btSubscriptionChannelGroupIds).addAll(btTargetUserGroupChannelGroupIds).build();
```
Dalamar42: Newlines between method calls in method chains

PR #428, opened 2016-10-12
New annotation for special BT channel groups

jamiesprax:
Added a new annotation for BT to use to get whitelisted generic_channel_groups_summary instead of altering the behaviour of the current channel_groups_summary annotation


atlas-api/src/main/java/org/atlasapi/query/v4/channel/ChannelWriter.java
```diff
@@ -83,7 +83,8 @@ public void write(@Nonnull ResolvedChannel entity, @Nonnull FieldWriter format,
         format.writeField("start_date", channel.getStartDate());
         format.writeField("advertised_from", channel.getAdvertiseFrom());
 
-        if (hasChannelGroupSummaryAnnotation(ctxt)) {
+        if (contextHasAnnotation(ctxt, Annotation.CHANNEL_GROUPS_SUMMARY) ||
```
Dalamar42: I think that's unnecessary. The way annotation implication should work is by adding the implied annotation in the active annotations

Dalamar42: Also you only added the new annotation to the channel group endpoint so it would never appear here

atlas-api/src/main/java/org/atlasapi/query/v4/channelgroup/ChannelGroupQueryExecutor.java
```diff
@@ -225,16 +225,18 @@ private ResolvedChannelGroup resolveAnnotationData(
                     Optional.absent()
         );
 
-        if(contextHasAnnotation(ctxt, Annotation.ADVERTISED_CHANNELS) ||
-                contextHasAnnotation(ctxt, Annotation.CHANNELS)) {
+        if (contextHasAnnotation(ctxt, Annotation.CHANNEL_GROUPS_SUMMARY)) {
```
Dalamar42: Ditto, if GENERIC_CHANNEL_GROUPS_SUMMARY is present you would also get CHANNEL_GROUPS_SUMMARY and therefore hit the first branch here and miss the second one you've added

atlas-api/src/main/java/org/atlasapi/query/v4/channelgroup/ChannelGroupQueryExecutor.java
```diff
@@ -225,16 +225,18 @@ private ResolvedChannelGroup resolveAnnotationData(
...
-                    contextHasAnnotation(ctxt, Annotation.CHANNEL_GROUPS_SUMMARY) ?
-                    resolveAdvertisedChannelsWithChannelGroups(channelGroup) :
-                    resolveAdvertisedChannels(channelGroup)
+                    resolveAdvertisedChannelsWithAllChannelGroups(channelGroup)
+            );
+        } else if (contextHasAnnotation(ctxt, Annotation.GENERIC_CHANNEL_GROUPS_SUMMARY)) {
+            resolvedChannelGroupBuilder.withAdvertisedChannels(
+                    resolveAdvertisedChannelsWithWhitelistedChannelGroups(channelGroup)
             );
-        } else if (contextHasAnnotation(ctxt, Annotation.CHANNEL_GROUPS_SUMMARY)) {
```
Dalamar42: Why `else if` here? is the `ADVERTISED_CHANNELS` annotation mutually exclusive with the others?

atlas-api/src/main/java/org/atlasapi/query/v4/channelgroup/ChannelGroupQueryExecutor.java
```diff
@@ -317,22 +307,43 @@ private ResolvedChannelGroup resolveAnnotationData(
...
+                                .withChannelGroupSummaries(
+                                        resolveAllChannelGroupSummaries(resolvedChannel.getChannel()))
+                                .build()
+                        )
+                        .collect(Collectors.toList())
         );
     }
 
-    private Optional<List<ChannelGroupSummary>> resolveChannelGroupSummaries(Channel channel) {
+    private Optional<List<ChannelGroupSummary>> resolveAllChannelGroupSummaries(
```
Dalamar42: There is a better way to do this to avoid code duplication. Change this method to accept a predicate, i.e. a function that accepts a broadcast Id and returns a boolean. Then in one case you pass:
```
broadcastId -> whitelistedIds.contains(broadcastId)
```

and in the other you pass:
```
broadcastId -> true
```

This way you avoid duplicating methods

PR #429, opened 2016-10-12
Ensure deleted broadcasts are removed from the content store

Dalamar42:
- As it were if a broadcast was removed we would never remove it from
  the store. Moreover if that broadcast was not actually removed, but
  only changed IDs we would end up having a duplicate broadcast, one
  with the old ID and one with the new one which would cause a
  schedule overlap
- Also removed the DatastaxCassandraContentStore which is not used
  and is superseded by the CqlContentStore


atlas-cassandra/src/main/java/org/atlasapi/content/AstyanaxProtobufContentMarshaller.java
```diff
@@ -15,18 +15,28 @@
...
+    }
+
     @Override
     protected void addColumnToBatch(ColumnListMutation<String> mutation, Id id, String column,
             byte[] value) {
         mutation.putColumn(column, value);
     }
 
     @Override
+    protected void removeColumn(ColumnListMutation<String> mutation, String column) {
```
emilssolmanis: Don't you just love it when someone extracts an interface that can only ever be implemented by an Astyanax driver using Protobuf? :D

Dalamar42: Oh yeah. Lots of :heart:

Dalamar42: For that matter, let me see if I can kill this and remove the extra layer of abstraction

PR #432, opened 2016-10-17
channel group summary whitelist by publisher

jamiesprax:
Using a hardcoded list of ID's would not catch new channel
groups coming in from the MPX ingester. Changing it to a
publisher whitelist means that any new channels will be
caught as they will have that publisher.

The old channel_groups_summary annotation has been
deprecated as it causes performance issues.


Dalamar42:
LGTM

PR #436, opened 2016-10-20
Fix bug where we write a broadcast with the wrong broadcast ID

Dalamar42:
- Before we were writing the broadcast ID based on the broadcast ID
  contained in the ScheduleUpdate which was delivered via messaging.
  However when we resolve the content currently in the DB for that
  broadcast we also try to match based on channel ID and transmission
  interval. That meant that if we had a broadcast that changed IDs
  since the schedule update message was generated we would end up
  writing that broadcast with the wrong broadcast ID clustering key.
  If that broadcast then became stale we would never be able to
  delete it as we would be trying to delete the wrong ID


Dalamar42:
Yeah, you are right. I'm adding one

PR #437, opened 2016-10-24
Add additional metrics

jamiesprax:
In GroupLock, metrics have been added to track how many threads
are waiting for locks and also a histogram of how many ids are
being locked at a time.

In the processing stores, metrics have been added to see how many
times they have been called, and failed.

In the workers, metrics have been added to see which publishers
are making the most requests.


atlas-api/src/main/java/org/atlasapi/AtlasPersistenceModule.java
```diff
@@ -129,6 +129,7 @@
     private static final ObjectMapper MAPPER = new ObjectMapper().findAndRegisterModules();
 
     private static final PersistenceAuditLog persistenceAuditLog = new NoLoggingPersistenceAuditLog();
+    private static final String METRIC_PREFIX = "persistence.util.";
```
Dalamar42: Only the `HashGenerator` should be using `persistence.util.`. The stores should be using `persistence.store.`

atlas-cassandra/src/main/java/org/atlasapi/CassandraPersistenceModule.java
```diff
@@ -61,6 +61,9 @@
 public class CassandraPersistenceModule extends AbstractIdleService implements PersistenceModule {
 
     private static final int CQL_PORT = 9042;
+
+    private static final String METRIC_PREFIX = "processing.store.";
```
Dalamar42: `persistence.store.`

atlas-cassandra/src/main/java/org/atlasapi/content/AstyanaxCassandraContentStore.java
```diff
@@ -184,6 +214,11 @@ private AstyanaxCassandraContentStore(
                 LongSerializer.get(), StringSerializer.get()
         );
         this.aliasIndex = AliasIndex.create(keyspace, cfName + "_aliases");
+
+        calledMeter = checkNotNull(metricRegistry)
```
Dalamar42: You can do the checkNotNulls once before you start using the parameters to avoid the duplicate check

atlas-cassandra/src/main/java/org/atlasapi/content/AstyanaxCassandraContentStore.java
```diff
@@ -223,6 +258,7 @@ private AstyanaxCassandraContentStore(
 
     @Override
     protected void doWriteContent(Content content, @Nullable Content previous) {
+        calledMeter.mark();
```
Dalamar42: These meters should only be in the public methods of the store which in this case I believe are in the abstract stores.

Also there should be a different meter for each method, not the same for all.

`persistence.store.<storeName>.<methodName>.etc`

atlas-cassandra/src/main/java/org/atlasapi/content/v2/CqlContentStore.java
```diff
@@ -139,116 +152,131 @@ public static Builder builder() {
...
-                    itemRef,
-                    ImmutableList.of(broadcastRefTranslator.serialize(broadcast.toRef()))
-            );
+    public void writeBroadcast(
+            ItemRef item,
+            Optional<ContainerRef> containerRef,
+            Optional<SeriesRef> seriesRef,
+            Broadcast broadcast
+    ) {
+        try {
```
Dalamar42: Missing called meter

atlas-cassandra/src/main/java/org/atlasapi/content/v2/CqlContentStore.java
```diff
@@ -801,8 +831,27 @@ public Builder withGraphStore(EquivalenceGraphStore val) {
             return this;
         }
 
+        public Builder withMetricRegistry(MetricRegistry val) {
+            metricRegistry = val;
```
Dalamar42: Don't use `val` as a variable name. IJ will autocomplete `metricRegistry` as the variable name which would be fine

atlas-core/src/main/java/org/atlasapi/util/GroupLock.java
```diff
@@ -20,32 +22,41 @@
...
 
-    private GroupLock(Ordering<? super T> ordering) {
+    private GroupLock(
+            Ordering<? super T> ordering,
+            MetricRegistry metricRegistry,
+            String metricPrefix
+    ) {
         this.ordering = ordering;
+
+        this.idsLocked = metricRegistry.histogram(metricPrefix + "histogram.idsLocked");
```
Dalamar42: `checkNotNull(metricPrefix)`

atlas-core/src/main/java/org/atlasapi/util/GroupLock.java
```diff
@@ -58,19 +69,10 @@ private GroupLock(Ordering<? super T> ordering) {
...
-                        "{} waiting on lock for {}",
-                        Thread.currentThread().getName(),
-                        id.toString()
-                );
-                locked.wait();
-            }
-            log.trace("{} acquired lock for {}", Thread.currentThread().getName(), id.toString());
-            locked.add(id);
-        }
+        threadsLocked.inc();
```
Dalamar42: For this it's better to do:

```
threadsLocked.inc();
try {
    // other stuff
} finally {
    threadsLocked.dec();
}
```

This way if an exception is thrown we still decrement the threadsLocked counter.

atlas-core/src/main/java/org/atlasapi/util/GroupLock.java
```diff
@@ -87,6 +89,7 @@ public void unlock(T id) {
         synchronized (locked) {
             if (locked.remove(id)) {
                 log.trace("{} unlocked {}", Thread.currentThread().getName(), id.toString());
+                idsLocked.update(-1);
```
Dalamar42: `idsLocked` is a histogram that records how many Ids are locked at a time every time someones calls lock(). It's not a counter. This statement effectively means that someone tried to lock -1 ids.

atlas-core/src/main/java/org/atlasapi/util/GroupLock.java
```diff
@@ -150,6 +157,7 @@ public void unlock(Set<T> ids) {
     public boolean tryLock(Set<T> ids) throws InterruptedException {
         synchronized (locked) {
             List<T> orderedIds = ordering.sortedCopy(ids);
+            idsLocked.update(ids.size());
```
Dalamar42: This doesn't have to happen inside the synchronized block. There is a performance cost to running things inside a synchronized (mutual exclusion) block because only one thread can run the code in there at a time so good practice is to only run the things that need to be synchronized in it.

atlas-core/src/main/java/org/atlasapi/util/GroupLock.java
```diff
@@ -150,6 +157,7 @@ public void unlock(Set<T> ids) {
...
             for (T id : orderedIds) {
                 if (!tryLock(id)) {
```
Dalamar42: The single ID `tryLock` is also public so we should do the same as for the single Id `lock`, i.e. create a separate `tryLockInternal` method and add metrics to `tryLock`

atlas-cassandra/src/main/java/org/atlasapi/event/ConcreteEventStore.java
```diff
@@ -35,13 +41,26 @@
...
+            String metricPrefix
+    ) {
         this.clock = checkNotNull(clock);
         this.idGenerator = checkNotNull(idGenerator);
         this.sender = checkNotNull(sender);
         this.eventHasher = checkNotNull(eventHasher);
         this.persistenceStore = checkNotNull(persistenceStore);
+
+        this.metricRegistry = metricRegistry;
+        write = metricPrefix + "write";
```
Dalamar42: Ditto missing trailing dot. This might be happening elsewhere as well. I haven't checked

atlas-cassandra/src/main/java/org/atlasapi/organisation/DatastaxCassandraOrganisationStore.java
```diff
@@ -72,6 +86,11 @@ protected DatastaxCassandraOrganisationStore(Session session,
                 .where(eq(PRIMARY_KEY_COLUMN, bindMarker(ORGANISATION_ID)))
                 .with(set(DATA_COLUMN, bindMarker(DATA))))
                 .setConsistencyLevel(writeConsistency);
+
+        this.metricRegistry = metricRegistry;
+
+        write = metricPrefix + "write";
```
Dalamar42: Trailing dot

atlas-cassandra/src/main/java/org/atlasapi/schedule/AbstractEquivalentScheduleStore.java
```diff
@@ -50,22 +51,41 @@
 
     private static final Logger log = LoggerFactory.getLogger(AbstractEquivalentScheduleStore.class);
 
+    private static final String METER_CALLED = ".meter.called";
```
Dalamar42: To avoid confusion it's best to be consistent with how we place dots in these strings. Other places place a trailing dot on the prefix rather than a leading one on the postfix so I think we should standardise on that.

PR #438, opened 2016-10-24
Mbst 17206/delete broadcasts with wrong ids

Dalamar42:


atlas-cassandra/src/main/java/org/atlasapi/schedule/CassandraEquivalentScheduleStore.java
```diff
@@ -383,21 +374,18 @@ public void updateContent(Iterable<Item> content) throws WriteException {
...
-                currentBroadcastRefs
-        );
 
-        ImmutableSet.Builder<StaleBroadcast> staleBroadcasts = ImmutableSet.builder();
+        ImmutableSet.Builder<BroadcastRow> staleBroadcasts = ImmutableSet.builder();
 
-        for (LocalDate day : currentBroadcastsByDay.keySet()) {
-            ImmutableCollection<BroadcastRef> currentBroadcastsInDay = currentBroadcastsByDay
+        for (LocalDate day : currentBroadcastRows.keySet()) {
+            ImmutableCollection<BroadcastRow> currentBroadcastsInDay = currentBroadcastRows
```
emilssolmanis: Dafuq, why not just iterate the `.entrySet()`?

atlas-cassandra/src/main/java/org/atlasapi/schedule/CassandraEquivalentScheduleStore.java
```diff
@@ -562,63 +550,76 @@ private ByteBuffer serialize(Broadcast broadcast) {
...
+        ImmutableMultimap.Builder<LocalDate, BroadcastRow> broadcastRows =
+                ImmutableMultimap.builder();
+
+        // We are getting the broadcastId from its column instead of the broadcast itself because
+        // we have seen situations where the broadcasts have been written under the wrong ID due
+        // to bugs. Using the broadcastId in the column guarantees we will either mark this row
+        // as stale or update it as appropriate
+        // We are also including rows with null broadcasts to ensure we give downstream code the
+        // chance to update or delete them
+        for (LocalDate day : futuresByDay.keySet()) {
```
emilssolmanis: Same, why not just `.entrySet()` it?

atlas-cassandra/src/main/java/org/atlasapi/schedule/CassandraEquivalentScheduleStore.java
```diff
@@ -562,63 +550,76 @@ private ByteBuffer serialize(Broadcast broadcast) {
...
+            )
+                    .stream()
+                    .flatMap(resultSet -> StreamSupport.stream(resultSet.spliterator(), false))
+                    .filter(row -> row.isNull(BROADCAST.name()) ||
+                            deserializeBroadcast(row).getTransmissionInterval().overlaps(interval))
+                    .map(row -> BroadcastRow.create(
+                            day,
+                            BROADCAST_ID.valueFrom(row)
+                    ))
+                    .forEach(broadcastRow -> broadcastRows.put(
```
emilssolmanis: Ouch. Why not just use a map collector?

Dalamar42: Because all these are MultiMaps and I don't want to process each pair separately. I want to process all the broadcast rows for each day in a batch.

In one of these cases I could change it to go over entries, but I think it would be less readable. In this one I would have to create a holder class to contain the day and also the processed rows. It would be quite messy.

emilssolmanis: I know, but don't we have a `MoreCollectors` for multimaps or something to that end? Either way, it's not a blocking thing.

Dalamar42: Yeah we do, but the problem here is in iterating over the multimap. Not collecting it.

PR #439, opened 2016-10-24
Change schedule equiv updates to not update the schedule

Dalamar42:
- Equivalence updates should only change the content in a given
  broadcast row, but never the schedule itself. This is done by
  -- not writing the broadcast when updating equivalences
  -- not resolving equiv schedule rows with null broadcasts
  -- resolving equiv schedule rows with null broadcasts as
  potentially stale when updating the schedule


atlas-cassandra/src/test/java/org/atlasapi/util/TestCassandraPersistenceModule.java
```diff
@@ -136,19 +135,17 @@ public void reset() throws ConnectionException {
         clearTables(session, context);
     }
 
+    public Session getSession() {
+        return cassandraService.getCluster().connect(keyspace);
```
emilssolmanis: Cache this?

PR #440, opened 2016-10-25
Fix debug endpoints showing Interval as millis

Dalamar42:


AdamDM:
lgtm

atlas-processing/src/main/java/org/atlasapi/system/debug/ContentDebugController.java
```diff
@@ -128,6 +125,19 @@ public ContentDebugController(
 
         this.jackson = new ObjectMapper();
         this.jackson.registerModule(new GuavaModule());
+
+        this.gson = new GsonBuilder()
+                .registerTypeAdapter(
+                        DateTime.class,
+                        (JsonSerializer<DateTime>) (src, typeOfSrc, context) ->
```
emilssolmanis: Btw I think we discovered that this one's unnecessary and just gets copy pasted along, but not 100%.

Dalamar42: Unfortunately I got a compilation error when I tried removing it.

PR #446, opened 2016-10-26
Mbst 17189/equiv schedule content updates

Dalamar42:


atlas-cassandra/src/main/java/org/atlasapi/schedule/AbstractEquivalentScheduleStore.java
```diff
@@ -152,39 +152,24 @@ public final void updateEquivalences(ImmutableSet<EquivalenceGraph> graphs)
...
 
+        ImmutableList<ResultSetFuture> futures = futureBuilder.build();
+
+        metricRegistry.histogram(updateEquivalences + "histogram.parallelWrites")
+                .update(futures.size());
+
         try {
             // Block until all futures are completed
-            for (CompletableFuture<Void> future : futures.build()) {
+            for (ResultSetFuture future : futures) {
```
emilssolmanis: These are all `ListenableFuture`s, you should be able to use `Futures.allAsList().get()`

For future -- the driver uses Guava quite heavily, so most of their future stuff is `Listenable*` and uses Guava's executors, etc.

PR #453, opened 2016-11-02
Change content/equivalence updates to not update rows older than 14 days

Dalamar42:
- This is intended to reduce the number of updates those workers have
  to do by accepting that the content and equivalence graph of old
  broadcasts may be a bit stale

### DIFF
--- a/atlas-cassandra/src/main/java/org/atlasapi/schedule/AbstractEquivalentScheduleStore.java
+++ b/atlas-cassandra/src/main/java/org/atlasapi/schedule/AbstractEquivalentScheduleStore.java
@@ -25,9 +25,11 @@ import org.atlasapi.media.entity.Publisher;
 
 import com.metabroadcast.common.collect.OptionalMap;
 import com.metabroadcast.common.stream.MoreCollectors;
+import com.metabroadcast.common.time.Clock;
 
 import com.codahale.metrics.MetricRegistry;
 import com.datastax.driver.core.ResultSetFuture;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Optional;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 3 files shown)


atlas-cassandra/src/main/java/org/atlasapi/schedule/AbstractEquivalentScheduleStore.java
```diff
@@ -50,6 +51,9 @@
 
     private static final Logger log = LoggerFactory.getLogger(AbstractEquivalentScheduleStore.class);
 
+    @VisibleForTesting
+    static final int MAX_BROADCAST_AGE_IN_DAYS_TO_UPDATE = 14;
```
emilssolmanis: Why not have this as an actual `Duration`?

atlas-cassandra/src/main/java/org/atlasapi/schedule/AbstractEquivalentScheduleStore.java
```diff
@@ -145,6 +148,11 @@ protected abstract void writeSchedule(
                 .collect(Collectors.toList());
     }
 
+    protected boolean shouldUpdateBroadcast(Broadcast broadcast) {
+        return broadcast.getTransmissionEndTime()
+                .isAfter(DateTime.now().minusDays(MAX_BROADCAST_AGE_IN_DAYS_TO_UPDATE));
```
emilssolmanis: Don't we have a `clock` here?

atlas-cassandra/src/test/java/org/atlasapi/schedule/CassandraEquivalentScheduleStoreIT.java
```diff
@@ -66,12 +72,9 @@ public static void cleanUp() throws Exception {
 
     @Test
     public void testWritingNewSchedule() throws Exception {
-
-        Channel channel = Channel.builder(Publisher.BBC).build();
-        channel.setId(1L);
         Interval interval = new Interval(
-                new DateTime(2014, 3, 21, 16, 0, 0, 0, DateTimeZones.UTC),
-                new DateTime(2014, 3, 21, 17, 0, 0, 0, DateTimeZones.UTC)
+                getDateTime(3, 21, 16, 0),
```
emilssolmanis: *make*DateTime?

PR #455, opened 2016-11-07
Rekey denormalisation without dates

emilssolmanis:


atlas-cassandra/src/main/java/org/atlasapi/content/v2/model/IdentifiedWithoutUpdateTimes.java
```diff
@@ -0,0 +1,37 @@
+package org.atlasapi.content.v2.model;
+
+import java.util.Set;
+
+import org.atlasapi.content.v2.model.udt.Alias;
+import org.atlasapi.content.v2.model.udt.Ref;
+
+/**
+ * @see Identified
```
Dalamar42: You can make these `{@link}` to those classes. IntelliJ is smart enough to refactor them then if we rename them or move them

atlas-cassandra/src/main/java/org/atlasapi/content/v2/model/udt/ItemRef.java
```diff
@@ -4,22 +4,21 @@
...
+/** This doesn't hold the actual ID and publisher because those are the strict PK of
+ * any resource ref. These objects are usually stored as a CQL {@code map<Ref, ItemRef>} and
+ * serialised accordingly. If you need to store a full {@code ItemRef} as a field, you'll need a
+ * to make a different UDT.
+ *
+ * @see Ref
+ * @see SeriesRef
+ */
 @UDT(name = "itemref")
 public class ItemRef {
```
Dalamar42: Is it worth naming this `PartialItemRef` or is that too dumb a name?

Dalamar42: To avoid conflicts if we need to add a full item ref UDT in the future

atlas-cassandra/src/test/java/org/atlasapi/content/v2/CqlContentStoreIT.java
```diff
@@ -44,10 +55,105 @@ protected ContentStore provideContentStore() {
     @Override
     public void testWritingResolvingContainerWhichOnlyChildRefsThrowsCorrectException() {}
 
-    @Ignore("MBST-17328")
     @Test
     @Override
     public void writingContentWithoutContainerRemovesExistingContainer() throws Exception {
```
Dalamar42: why did you need to rewrite this test from what's in the abstract class?

atlas-cassandra/src/main/java/org/atlasapi/content/v2/model/IdentifiedWithoutUpdateTimes.java
emilssolmanis: eh? `@see` already produces a link, `{@link}` is for links inside the paragraphs.

atlas-cassandra/src/test/java/org/atlasapi/content/v2/CqlContentStoreIT.java
emilssolmanis: Because I thought you said you didn't fix it in the Asty store and so there was no test? Mostly because I'm an idiot and wrote this at 7 pm on a Friday I suppose.

PR #456, opened 2016-11-11
Add support for serving schedules from PA & EBS to BT

jamiesprax:
* Application access role added to determine whether functionality is
needed for the requesting application

* "source" field added to schedule output for channels to show where
the source of the channel data

* Application access role checking in the executor resolves channels from
EBS if that is the applications preferred source

### DIFF
--- a/atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleQueryExecutorImpl.java
+++ b/atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleQueryExecutorImpl.java
@@ -1,13 +1,14 @@
 package org.atlasapi.query.v4.schedule;
 
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.List;
-import java.util.concurrent.TimeUnit;
-import java.util.stream.Collectors;
-import java.util.stream.StreamSupport;
-
-import org.apache.avro.generic.GenericData;
+import com.google.common.base.Function;
+import com.google.common.base.Functions;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
+import com.google.common.util.concurrent.AsyncFunction;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
 import org.atlasapi.annotation.Annotation;
 import org.atlasapi.application.ApplicationAccessRole;
 import org.atlasapi.application.ApplicationSources;
@@ -30,19 +31,12 @@ import org.atlasapi.query.common.QueryResult;
 import org.atlasapi.schedule.ChannelSchedule;
 import org.atlasapi.schedule.Schedule;
 import org.atlasapi.schedule.ScheduleResolver;
-
-import com.metabroadcast.common.stream.MoreCollectors;
-
-import com.google.common.base.Function;
-import com.google.common.base.Functions;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Iterables;
-import com.google.common.collect.Lists;
-import com.google.common.util.concurrent.AsyncFunction;
-import com.google.common.util.concurrent.Futures;
-import com.google.common.util.concurrent.ListenableFuture;
 import org.joda.time.Interval;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
--- a/atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleQueryExecutorImpl.java
+++ b/atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleQueryExecutorImpl.java
@@ -2,10 +2,7 @@ package org.atlasapi.query.v4.schedule;
 
 import com.google.common.base.Function;
 import com.google.common.base.Functions;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Iterables;
-import com.google.common.collect.Lists;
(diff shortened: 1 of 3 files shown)


atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleQueryExecutorImpl.java
```diff
@@ -63,38 +66,46 @@ public ScheduleQueryExecutorImpl(ChannelResolver channelResolver,
...
+            defaultChannels.removeAll(ebsChannels);
+        }
+
+        List<ChannelSchedule> channelScheduleList = getChannelScheduleList(ebsChannels, defaultChannels, query);
 
         if (query.isMultiChannel()) {
-            List<ChannelSchedule> channelSchedules = channelSchedules(schedule, query);
             return QueryResult.listResult(
-                    channelSchedules,
+                    channelScheduleList,
```
Dalamar42: Don't name the variable by the type because the type could change in the future. `channelSchedules` is enough for this

atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleQueryExecutorImpl.java
```diff
@@ -63,38 +66,46 @@ public ScheduleQueryExecutorImpl(ChannelResolver channelResolver,
...
+
+        List<Channel> defaultChannels = Lists.newArrayList(channels);
+        List<Channel> ebsChannels = new ArrayList<>();
+
+        if (query.getContext().getApplicationSources().getAccessRoles().contains(
+                ApplicationAccessRole.PREFER_EBS_SCHEDULE)) {
+
+            ebsChannels = defaultChannels.stream()
+                    .filter(channel -> channel.getAvailableFrom().contains(Publisher.BT_SPORT_EBS))
+                    .collect(Collectors.toList());
```
Dalamar42: Nitpick. Both channel lists could be made immutable, given they are being passed to another class

atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleQueryExecutorImpl.java
```diff
@@ -63,38 +66,46 @@ public ScheduleQueryExecutorImpl(ChannelResolver channelResolver,
...
     }
 
     private Iterable<Channel> resolveChannels(ScheduleQuery query) throws QueryExecutionException {
-        Iterable<Id> channelIds;
-        if (query.isMultiChannel()) {
-            channelIds = query.getChannelIds();
-        } else {
-            channelIds = ImmutableSet.of(query.getChannelId());
-        }
+        Iterable<Id> channelIds = query.isMultiChannel() ?
```
Dalamar42: Code style for ternary is:
```
query.isMultichannel()
        ? query.getChannelIds()
        : ImmutableSet.of(query.getChannelId());
```

atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleQueryExecutorImpl.java
```diff
@@ -149,34 +154,62 @@ public ScheduleQueryExecutorImpl(ChannelResolver channelResolver,
...
+            replaceBroadcasts(item, input.getBroadcast());
+            return new ItemAndBroadcast(item, input.getBroadcast());
         });
     }
 
     private void replaceBroadcasts(Item item, Broadcast broadcast) {
         item.setBroadcasts(ImmutableSet.of(broadcast));
     }
+
+    private List<ChannelSchedule> getChannelScheduleList(
```
Dalamar42: Ditto, `getChannelSchedules` and return ImmutableList

atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleQueryExecutorImpl.java
```diff
@@ -149,34 +154,62 @@ public ScheduleQueryExecutorImpl(ChannelResolver channelResolver,
...
+            List<Channel> ebsChannels,
+            List<Channel> defaultChannels,
+            ScheduleQuery query
+    ) throws ScheduleQueryExecutionException {
+
+        List<ChannelSchedule> ebsChannelSchedule = new ArrayList<>();
+        List<ChannelSchedule> defaultChannelSchedule = new ArrayList<>();
+
+        if (!ebsChannels.isEmpty()) {
```
Dalamar42: Don't think you need to check if it's empty here. Check the `scheduleResolver.resolve` method, but it should handle an empty channel list gracefully

atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleQueryExecutorImpl.java
```diff
@@ -63,38 +66,46 @@ public ScheduleQueryExecutorImpl(ChannelResolver channelResolver,
...
         }
-        ListenableFuture<Schedule> schedule = scheduleResolver.resolve(
-                channels,
-                new Interval(query.getStart(), query.getEnd().get()),
-                query.getSource()
-        );
+
+        Iterable<Channel> channels = resolveChannels(query);
+
+        List<Channel> defaultChannels = Lists.newArrayList(channels);
```
Dalamar42: Given the other bug with channel ordering we need to make sure all channels stay in the same order regardless of how they are resolved.

You can make a inner class that contains a channel and a source to represent them and then pass that around as a single list

atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleQueryExecutorImpl.java
```diff
@@ -149,34 +154,62 @@ public ScheduleQueryExecutorImpl(ChannelResolver channelResolver,
...
+            ListenableFuture<Schedule> baseSchedule = scheduleResolver.resolve(
+                    defaultChannels,
+                    new Interval(query.getStart(), query.getEnd().get()),
+                    query.getSource()
+            );
+
+            defaultChannelSchedule = channelSchedules(baseSchedule, query);
+        }
+
+        return ImmutableList.<ChannelSchedule>builder()
```
Dalamar42: Ditto we need to make sure we keep the channel order.

PR #464, opened 2016-11-19
Add dynamically generated graph blacklist

Dalamar42:
- The blacklist contains a list of graphs to be blocked and is
  updated with all the nodes in that graph every time it is
  encountered. Over time it should grow to contain all IDs in a
  snapshot of the given graphs

### DIFF
--- a/atlas-core/src/main/java/org/atlasapi/equivalence/AbstractEquivalenceGraphStore.java
+++ b/atlas-core/src/main/java/org/atlasapi/equivalence/AbstractEquivalenceGraphStore.java
@@ -6,6 +6,7 @@ import java.util.Queue;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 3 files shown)


atlas-core/src/main/java/org/atlasapi/equivalence/AbstractEquivalenceGraphStore.java
```diff
@@ -148,28 +164,30 @@ public AbstractEquivalenceGraphStore(
...
+                updateBlacklist(graphUpdate);
+
                 sendUpdateMessage(subject, graphUpdate);
             }
             return updated;
 
         } catch (InterruptedException e) {
             metricRegistry.meter(updateEquivalences + METER_FAILURE).mark();
-            log().error(String.format("%s: %s", subject, newAdjacents), e);
+            log.error(String.format("%s: %s", subject, newAdjacents), e);
```
emilssolmanis: This can be `log.error("{}: {}", subject, newAdjacents, e)` these days. It's a hack for an old version of slf4j that didn't check for `instanceof Exception` on the last arg in `.error(String, Object...)`.

emilssolmanis: Same for the other places that try to explicitly stick to `.error(String, Exception)`.

PR #465, opened 2016-11-21
Feature/abstract equiv graph store

Dalamar42:


atlas-core/src/main/java/org/atlasapi/equivalence/AbstractEquivalenceGraphStore.java
```diff
@@ -229,13 +237,17 @@ private void sendUpdateMessage(ResourceRef subject, EquivalenceGraphUpdate updat
                     Thread.currentThread().getName()
             );
 
+            int lockAttempts = 0;
```
emilssolmanis: Given the way this is going, might wanna make that a `long`...

PR #467, opened 2016-11-29
Mbst 17229/serving schedules multiple sources

jamiesprax:
This will allow us to display the source the channel schedule was resolved from.


atlas-core/src/main/java/org/atlasapi/schedule/ChannelSchedule.java
```diff
@@ -16,27 +17,29 @@
...
+    private final Publisher source;
 
     public ChannelSchedule(Channel channel, Interval interval, Iterable<ItemAndBroadcast> entries) {
         this.channel = checkNotNull(channel);
         this.interval = checkNotNull(interval);
         this.entries = Ordering.natural().immutableSortedCopy(entries);
+        this.source = null;
+    }
+
+    private ChannelSchedule(Channel channel, Interval interval, Iterable<ItemAndBroadcast> entries, Publisher source) {
```
AdamDM: Use a static create method, maybe even a builder as it has 4 parameters

PR #468, opened 2016-11-29
removed internal config from dockerfile

lhopki01:


emilssolmanis:
Same question as before, where's all this going?

lhopki01:
@emilssolmanis It's all going into the configmaps on kubernetes.

PR #469, opened 2016-11-30
Move multiple schedule logic to correct executor

jamiesprax:
* Moved logic to correct executor - EquivalentScheduleQueryExecutor
* Reverted old executor back to pre-feature state - ScheduleQueryExecutorImpl
* Added schedule source to ChannelSchedule
* Fixed schedule fallback bug if EBS schedule fails to be resolved

### DIFF
--- a/atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleQueryExecutorImpl.java
+++ b/atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleQueryExecutorImpl.java
@@ -1,12 +1,9 @@
 package org.atlasapi.query.v4.schedule;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
-import java.util.stream.StreamSupport;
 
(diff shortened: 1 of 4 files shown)


atlas-api/src/main/java/org/atlasapi/query/v4/schedule/ScheduleQueryExecutorImpl.java
```diff
@@ -70,45 +63,39 @@ public ScheduleQueryExecutorImpl(ChannelResolver channelResolver,
...
         );
     }
 
     private Iterable<Channel> resolveChannels(ScheduleQuery query) throws QueryExecutionException {
-        Iterable<Id> channelIds = query.isMultiChannel()
-                ? query.getChannelIds()
-                : ImmutableSet.of(query.getChannelId());
-
+        Iterable<Id> channelIds;
+        if (query.isMultiChannel()) {
```
AdamDM: can use a ternary conditional operator here

AdamDM:
olgtm

PR #471, opened 2016-12-02
Fix for Deer not returning channels in query order

jamiesprax:
The order of the original query is now used to order the channel
schedules when they have all been resolved.
Previously, the order of the resolved channels was used which was
not the same.


AdamDM:
lgtm

PR #478, opened 2016-12-05
Feature/prometheus metrics

Dalamar42:
Last commit only please

### DIFF
--- a/atlas-api/src/main/java/org/atlasapi/system/MetricsController.java
+++ b/atlas-api/src/main/java/org/atlasapi/system/MetricsController.java
@@ -1,0 +1,33 @@
+package org.atlasapi.system;
+
+import java.io.IOException;
+
+import javax.servlet.ServletException;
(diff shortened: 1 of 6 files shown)


AdamDM:
lgtm

atlas-api/src/main/java/org/atlasapi/system/MetricsController.java
```diff
@@ -0,0 +1,33 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import io.prometheus.client.CollectorRegistry;
+import io.prometheus.client.exporter.MetricsServlet;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+
+@Controller()
```
emilssolmanis: The `()` are a bit confusing and unnecessary.

PR #480, opened 2016-12-09
Cope with missing containers in content bootstrap worker

emilssolmanis:


atlas-processing/src/main/java/org/atlasapi/system/bootstrap/workers/ContentBootstrapWorker.java
```diff
@@ -120,8 +121,31 @@ private void process(Id contentId, Timestamp timestamp) throws WriteException {
...
+                Id containerId = mre.getMissingId();
+                log.info("Missing container {} for {}, trying to fix", containerId, contentId);
+
+                Resolved<Content> missing = resolveContent(containerId);
+                if (missing.getResources().isEmpty()) {
+                    throw e;
+                }
+
+                writer.writeContent(missing.getResources().first().get());
+                result = writer.writeContent(contentToWrite);
```
Dalamar42: So if this throws an exception we are ok to ignore it. It's a bit smelly, but given what we currently do in the catch above I think that's reasonable.

atlas-processing/src/main/java/org/atlasapi/system/bootstrap/workers/ContentBootstrapWorker.java
```diff
@@ -120,8 +121,31 @@ private void process(Id contentId, Timestamp timestamp) throws WriteException {
...
 
-        WriteResult<Content, Content> result =
-                writer.writeContent(content.getResources().first().get());
+        WriteResult<Content, Content> result = null;
+
+        Content contentToWrite = content.getResources().first().get();
+
+        try {
+            result = writer.writeContent(contentToWrite);
+        } catch (WriteException e) {
```
Dalamar42: Obvious comment: Programme flow via exceptions...
You don't have much of a choice here though and it is a temporary bootstrap so we should be able to live with it.

atlas-cassandra/src/main/java/org/atlasapi/content/v2/CqlContentStore.java
```diff
@@ -318,6 +316,19 @@ public void writeBroadcast(
...
+    @Nullable
+    private Content deserializeIfFull(@Nullable org.atlasapi.content.v2.model.Content content) {
+        // we denormalise things like itemRefs on containers. When a content gets written, we write
+        // its itemRef onto the container without checking that the container exists, thus
+        // creating an incomplete container object. These can be recognised most notably by
+        // lack of 'type'
+        if (content != null && content.getType() != null) {
+            return translator.deserialize(content);
+        } else {
+            return null;
```
Dalamar42: This is very annoying, but I don't see a way around it short of rewriting the store's interface...

PR #481, opened 2016-12-12
Fix ClassCastException in kafka queues

jamiesprax:
The same method was being used to add locations for nested Aliases which are returned as maps from ES client instead of Aliases like the top level ones.
A new implementation for nested aliases has been added to map them to EsAliases.


atlas-elasticsearch/src/main/java/org/atlasapi/content/EsLocation.java
```diff
@@ -66,8 +80,17 @@ public static EsLocation fromMap(Map<String, Object> map) {
         }
 
         if (map.get(ALIASES) != null) {
-            esLocation.aliases((Iterable<Alias>)map.get(ALIASES));
+            if(((Iterable)map.get(ALIASES)).iterator().next().getClass() != Alias.class) {
```
Dalamar42: This check is unnecessary and a little misleading. It implies that sometimes we may get an `Iterable<Alias>` back which can never happen.

atlas-elasticsearch/src/main/java/org/atlasapi/content/EsLocation.java
```diff
@@ -66,8 +80,17 @@ public static EsLocation fromMap(Map<String, Object> map) {
...
+                esLocation.aliasesFromEs((List<EsAlias>)
```
Dalamar42: You shouldn't need this cast here. Java should be able to infer the type from the stream code

atlas-elasticsearch/src/main/java/org/atlasapi/content/EsLocation.java
```diff
@@ -66,8 +80,17 @@ public static EsLocation fromMap(Map<String, Object> map) {
...
+                        ((List<Map>)map.get(ALIASES)).stream()
+                        .map(EsAlias::fromMap)
```
Dalamar42: Indentation

atlas-elasticsearch/src/main/java/org/atlasapi/content/EsLocation.java
```diff
@@ -66,8 +80,17 @@ public static EsLocation fromMap(Map<String, Object> map) {
...
+                        .collect(Collectors.toList())
+                );
+            } else {
+                esLocation.aliases((Iterable)map.get(ALIASES));
```
Dalamar42: Ditto as above. Unless I'm misunderstanding the code, this can never happen

atlas-elasticsearch/src/main/java/org/atlasapi/util/EsAlias.java
```diff
@@ -12,6 +13,10 @@
 
 public class EsAlias extends EsObject {
 
+    private static final Function<Alias, EsAlias> TO_ALIAS = EsAlias::valueOf;
```
Dalamar42: 👍 for getting rid of that nonsense with this

atlas-elasticsearch/src/test/java/org/atlasapi/content/EsLocationTest.java
```diff
@@ -49,10 +52,23 @@ public void setUp() {
...
-        map.put(AVAILABILITY_TIME, formatter.parseDateTime(DATE_A));
-        map.put(AVAILABILITY_END_TIME, formatter.parseDateTime(DATE_B));
-        map.put(ALIASES, aliases.stream().collect(MoreCollectors.toImmutableSet()));
+        aliasMap = new HashMap<>();
+        aliasMap.put(AVAILABILITY_TIME, formatter.parseDateTime(DATE_A));
+        aliasMap.put(AVAILABILITY_END_TIME, formatter.parseDateTime(DATE_B));
+        aliasMap.put(ALIASES, aliases.stream().collect(MoreCollectors.toImmutableSet()));
+
+        Map<String, Object> aliasesAsHashMap = Maps.newHashMap();
+        Map<String, String> aliasAMap = Maps.newHashMap();
```
Dalamar42: You can use:
```
ImmutableMap.of(
    "keyA", "valueA",
    "keyB", "valueB"
);
```
More succinct and readable that way.

atlas-elasticsearch/src/main/java/org/atlasapi/content/EsLocation.java
jamiesprax: If i remove it, it complains that it can't cast it but then if i put it in, it marks it grey and says it can be inferred. The first case was a compilation error so I had to include it

PR #482, opened 2016-12-13
Filter unpublished broadcasts when denormalising into equiv store

emilssolmanis:


atlas-cassandra/src/test/java/org/atlasapi/schedule/CassandraEquivalentScheduleStoreIT.java
```diff
@@ -2082,6 +2086,65 @@ public void updatingEquivalencesDoesNotUpdateOldBroadcastRows() throws Exception
         assertThat(equivalenceSet.contains(item.getId()), is(true));
     }
 
+    @Test
+    public void unpublishedBroadcastsAreExcludedWhenLookingForDenormalisation() throws Exception {
```
Dalamar42: I'm guessing changing the code to use a LinkedHashSet or something is not reasonable?

emilssolmanis: @Dalamar42 Means a model change just for testing. I'm not necessarily against it, but it's a new contract we're adding there.

PR #483, opened 2016-12-14
Add equivalence graph filtering to equiv content/schedule stores

Dalamar42:
- Content should now only resolve if:
  -- The content is actively published
  -- The content's source is enabled in the application
  -- The content is reachable on the equivalence graph through other
     content whose source is enabled in the application
  The entry point for graph traversal in the content store is the
  requested ID and in the schedule store it's the ID of the
  broadcast's item


atlas-cassandra/src/main/resources/atlas.schema
```diff
@@ -124,6 +124,7 @@ CREATE TABLE IF NOT EXISTS equivalent_schedule (
   content blob,
   schedule_update timestamp,
   equiv_update timestamp,
+  broadcast_item_id bigint,
```
emilssolmanis: Don't forget that this will need an `ALTER TABLE` before deployment.

PR #484, opened 2016-12-15
Fix/class cast exception on new locations

jamiesprax:
New locations were converted to EsObjects and then added to the list
of existing locations which were still Maps. This caused an additional casting conflict.
Instead of mapping the new locations to EsObjects, they are now translated
to maps to be mapped to EsObjects later on with the existing ones.

### DIFF
--- a/atlas-elasticsearch/src/main/java/org/atlasapi/content/EsContentTranslator.java
+++ b/atlas-elasticsearch/src/main/java/org/atlasapi/content/EsContentTranslator.java
@@ -35,6 +35,7 @@ import org.elasticsearch.action.get.GetRequest;
 import org.elasticsearch.action.get.GetResponse;
 import org.elasticsearch.client.Client;
 import org.elasticsearch.client.Requests;
+import org.elasticsearch.common.collect.Maps;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


atlas-elasticsearch/src/main/java/org/atlasapi/content/EsContentTranslator.java
```diff
@@ -360,8 +362,7 @@ private Object dedupeAndMergeTransmissionTime(Item episode, List<Integer> list)
 
         ImmutableList<Map<String, Object>> newLocations = locations.stream()
                 .filter(location -> filter.apply(location.getPolicy()))
-                .map(this::toEsLocation)
```
Dalamar42: I am a bit confused about the fix. This seems to already be doing what the commit message is describing.

atlas-elasticsearch/src/main/java/org/atlasapi/content/EsContentTranslator.java
```diff
@@ -360,8 +362,7 @@ private Object dedupeAndMergeTransmissionTime(Item episode, List<Integer> list)
...
-                .map(EsObject::toMap)
+                .map(location -> newLocationPolicyToMap(location.getPolicy()))
```
Dalamar42: This needs to be a map of locations, but here we are only converting the policy and ignoring the rest of the location.

atlas-elasticsearch/src/main/java/org/atlasapi/content/EsContentTranslator.java
```diff
@@ -370,6 +371,22 @@ private Object dedupeAndMergeTransmissionTime(Item episode, List<Integer> list)
...
 
+    private Map<String, Object> newLocationPolicyToMap(Policy policy) {
+        ImmutableList.Builder<Map<String, String>> aliases = ImmutableList.builder();
+
+        policy.getAliases().forEach(alias -> aliases.add(ImmutableMap.of(
+                EsAlias.NAMESPACE, alias.getNamespace(),
+                EsAlias.VALUE, alias.getValue())
+        ));
+
+        return ImmutableMap.of(
```
Dalamar42: The conversation to a map should be encapsulated in EsLocation. Isn't there a method for this already?

atlas-elasticsearch/src/main/java/org/atlasapi/content/EsLocation.java
```diff
@@ -80,13 +84,29 @@ public static EsLocation fromMap(Map<String, Object> map) {
...
+            esLocation.aliasesFromEs(
+                    (List<EsAlias>) ((List<Map>) map.get(ALIASES)).stream()
+                            .map(EsAlias::fromMap)
+                            .collect(Collectors.toList())
+            );
         }
         return esLocation;
     }
 
+    public static Map<String, Object> toMap(Location location) {
```
Dalamar42: For consistency with the rest of the code this method should accept an `EsLocation`, not a Location

atlas-elasticsearch/src/main/java/org/atlasapi/content/EsContentTranslator.java
```diff
@@ -360,8 +362,7 @@ private Object dedupeAndMergeTransmissionTime(Item episode, List<Integer> list)
 
         ImmutableList<Map<String, Object>> newLocations = locations.stream()
                 .filter(location -> filter.apply(location.getPolicy()))
-                .map(this::toEsLocation)
-                .map(EsObject::toMap)
+                .map(EsLocation::toMap)
```
Dalamar42: for consistency this should be
```
.map(this::toEsLocation)
.map(EsLocation::toMap)
```

atlas-elasticsearch/src/main/java/org/atlasapi/content/EsLocation.java
Dalamar42: This maintains the style of other `EsX` objects that they only deal with themselves and the map representation and it's the translator class that needs to know how to convert a `Location` to an `EsLocation` and vv.

PR #487, opened 2016-12-20
Move delphi /adjacents/{id}.json endpoint to Atlas Deer app servers

Dalamar42:
- Also rename to /delphi/graph/{id}.json


atlas-processing/src/main/java/org/atlasapi/system/debug/serializers/EquivalenceGraphSerializer.java
```diff
@@ -0,0 +1,58 @@
...
+        json.add(
+                "id",
+                context.serialize(codec.encode(BigInteger.valueOf(src.getId().longValue())))
+        );
+        json.add(
+                "updated",
+                context.serialize(src.getUpdated())
+        );
+
+		JsonArray jsonArray = new JsonArray();
```
emilssolmanis: Is that... a **TAB**?!?

fubar

PR #493, opened 2017-01-09
Mbst 17626/query parsing clean up

Dalamar42:


atlas-api/src/main/java/org/atlasapi/query/common/coercers/BooleanCoercer.java
```diff
@@ -0,0 +1,52 @@
...
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
+
+public class BooleanCoercer implements AttributeCoercer<Boolean> {
+
+    private final ImmutableSet<String> validInput =
+            ImmutableSet.copyOf(
+                    Lists.transform(
+                            ImmutableList.of(Boolean.TRUE, Boolean.FALSE),
```
emilssolmanis: Wtf. This can easily be expressed as a single stream with 2 `.map()`s avoiding all this crazy nesting.

atlas-api/src/main/java/org/atlasapi/query/common/coercers/FloatCoercer.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    public List<Float> apply(Iterable<String> values) throws InvalidAttributeValueException {
+        ImmutableList.Builder<Float> floats = ImmutableList.builder();
+
+        for (String value : values) {
+            try {
+                floats.add(Float.valueOf(value));
+            } catch (NumberFormatException nfe) {
+                throw new InvalidAttributeValueException(value);
```
emilssolmanis: Not to be a little female dog, but generally good style to include the cause in the exception you throw.

atlas-api/src/main/java/org/atlasapi/query/common/coercers/BooleanCoercer.java
Dalamar42: Yeah I saw this, but I wasn't sure how much I wanted to change here. I'll sort it out

atlas-api/src/main/java/org/atlasapi/query/common/coercers/FloatCoercer.java
Dalamar42: Copy pasta. Will fix

PR #495, opened 2017-01-12
Feature/kafka rollback replayer endpoint

emilssolmanis:
Assumes we have the ability to replay a topic from the start, which we don't at the moment. Change in `common-queue` to follow, since we've needed this much more than once by now.

### DIFF
--- a/atlas-processing/src/main/java/org/atlasapi/system/bootstrap/BootstrapModule.java
+++ b/atlas-processing/src/main/java/org/atlasapi/system/bootstrap/BootstrapModule.java
@@ -78,20 +78,20 @@ public class BootstrapModule {
 
     @Bean
     ContentBootstrapController contentBootstrapController() {
-        return new ContentBootstrapController(
-                persistence.legacyContentResolver(),
-                persistence.legacyContentLister(),
-                persistence.nullMessageSendingContentStore(),
-                search.equivContentIndex(),
-                explicitEquivalenceMigrator,
-                NUMBER_OF_SOURCE_BOOTSTRAP_TRHEADS,
-                progressStore(),
-                metrics,
-                persistence.nullMessageSendingEquivalentContentStore(),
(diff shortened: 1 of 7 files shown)


atlas-processing/src/main/java/org/atlasapi/system/bootstrap/BootstrapModule.java
```diff
@@ -78,20 +78,20 @@ BootstrapController bootstrapController() {
...
-                persistence.contentStore(),
-                persistence.neo4jContentStore()
-        );
+        return ContentBootstrapController.builder()
+                .withRead(persistence.legacyContentResolver())
+                .withContentLister(persistence.legacyContentLister())
+                .withWrite(persistence.nullMessageSendingContentStore())
+                .withContentIndex(search.equivContentIndex())
+                .withEquivalenceMigrator(explicitEquivalenceMigrator)
+                .withMaxSourceBootstrapThreads(NUMBER_OF_SOURCE_BOOTSTRAP_TRHEADS)
```
AdamDM: maybe rename TRHEADS while you're in there

AdamDM:
olgtm

atlas-processing/src/main/java/org/atlasapi/system/bootstrap/ContentBootstrapController.java
```diff
@@ -203,7 +222,79 @@ public void bootstrapContent(
...
+            HttpServletResponse response
+    ) {
+        if (this.replayBootstrapWorker != null) {
+            response.setStatus(HttpStatus.BAD_REQUEST.value());
+            return;
+        }
+
+        DateTime since = DateTime.parse(dateTimeString);
+
+        this.replayBootstrapWorker = replayConsumerFactory.apply(
```
Dalamar42: A bit of a naming thing. We call workers only the classes that process the the messages. The actual service we call listeners.

PR #496, opened 2017-01-12
Add metrics around equiv item filter logic

Dalamar42:
- This filtering logic currently has a bug (MBST-17731) in that it
  removes items it should not. These metrics are intended to give
  some visibility into how much extra work Cassandra will have to do
  after this is fixed

### DIFF
--- a/atlas-cassandra/src/main/java/org/atlasapi/schedule/AbstractEquivalentScheduleStore.java
+++ b/atlas-cassandra/src/main/java/org/atlasapi/schedule/AbstractEquivalentScheduleStore.java
@@ -35,6 +35,7 @@ import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Multimaps;
 import com.google.common.util.concurrent.Futures;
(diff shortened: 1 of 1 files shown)


atlas-cassandra/src/main/java/org/atlasapi/schedule/AbstractEquivalentScheduleStore.java
```diff
@@ -354,25 +387,28 @@ private String getLockKey(Id channelId, Publisher source, LocalDate date) {
...
             );
 
             if (bestMatch.isPresent()) {
                 Item item = bestMatch.get().copy();
-                item.setBroadcasts(matchingOrEmpty(subjectBroadacast, item.getBroadcasts()));
+                item.setBroadcasts(matchingOrEmpty(subjectBroadcast, item.getBroadcasts()));
                 selected.add(item);
             } else {
                 selected.addAll(matchingOrEmptyBroadcasts(
-                        subjectBroadacast, sourceContent.getValue()
```
emilssolmanis: But what if that was supposed to be an Italian accent...? 

luigi

atlas-cassandra/src/main/java/org/atlasapi/schedule/AbstractEquivalentScheduleStore.java
```diff
@@ -49,7 +50,8 @@
 
 public abstract class AbstractEquivalentScheduleStore implements EquivalentScheduleStore {
 
-    private static final Logger log = LoggerFactory.getLogger(AbstractEquivalentScheduleStore.class);
+    private static final Logger log = LoggerFactory.getLogger(AbstractEquivalentScheduleStore
+            .class);
```
AdamDM: private static final Logger log = LoggerFactory.getLogger(
                                        AbstractEquivalentScheduleStore.class
                               );

Would this look more normal ^ ?

AdamDM:
olgtm

atlas-cassandra/src/main/java/org/atlasapi/schedule/AbstractEquivalentScheduleStore.java
Dalamar42: Automatic reformat fail. Fixing

PR #497, opened 2017-01-13
Fix schedule store to not filter items by source

Dalamar42:
- Previously when storing a schedule entry (single broadcast) we:
  -- Kept all items for each source in the equiv graph if none of the
     broadcasts of these items matched the broadcast of the schedule
     entry
  -- Kept only one item per source if one such item was found that
     matched the broadcast of the schedule entry
- This logic is contrary to what we expect here which is that every
  schedule row will have the entire equivalence graph. This commit
  fixes this behaviour.
- We are preserving the logic that filters out any non-matching
  broadcasts since they are not relevant in the context of the
  equivalent schedule store

### DIFF
--- a/atlas-cassandra/src/main/java/org/atlasapi/schedule/AbstractEquivalentScheduleStore.java
+++ b/atlas-cassandra/src/main/java/org/atlasapi/schedule/AbstractEquivalentScheduleStore.java
@@ -15,7 +15,6 @@ import org.atlasapi.content.ContentResolver;
 import org.atlasapi.content.Item;
 import org.atlasapi.entity.Id;
 import org.atlasapi.entity.Identifiables;
-import org.atlasapi.entity.Sourceds;
 import org.atlasapi.entity.util.WriteException;
 import org.atlasapi.equivalence.EquivalenceGraph;
 import org.atlasapi.equivalence.EquivalenceGraphStore;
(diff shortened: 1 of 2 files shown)


atlas-cassandra/src/main/java/org/atlasapi/schedule/AbstractEquivalentScheduleStore.java
```diff
@@ -384,74 +348,31 @@ private String getLockKey(Id channelId, Publisher source, LocalDate date) {
...
                     return copy;
                 })
-                .collect(Collectors.toList());
+                .collect(MoreCollectors.toImmutableSet());
     }
 
-    private java.util.Optional<Item> bestMatch(
-            Collection<Item> sourceContent,
-            Broadcast subjectBroadcast
+    private Set<Broadcast> filterToOnlyMatchingBroadcasts(
```
emilssolmanis: Not convinced about the naming. It's plural, but the method does a `.findFirst()`, by definition it will only ever return 1 item.

Dalamar42: The signature is confusing yeah. It is only meant to return either one broadcast or none, but it returns a set. I think I'll rename this and move the set transformation to the caller to make it obvious what's happening here.

PR #503, opened 2017-01-17
use source writer with ID's

jamiesprax:
The writer that was being used by /system/sources.json endpoint did not match the one used in  /4/admin/sources.json


dragospitica:
LGTM

PR #505, opened 2017-01-19
Mbst 17375/migrate to applications service client

jamiesprax:
* Implemented applications service client
* Removed old Application & ApplicationSources
* Removed Auth, Users, and Sources from Deer (no longer used)
* Updated tests


atlas-api/src/main/java/org/atlasapi/application/ApiKeyApplicationFetcher.java
```diff
@@ -0,0 +1,69 @@
...
+    @VisibleForTesting
+    ApiKeyApplicationFetcher(ApplicationsClient applicationsClient, Environment environment) {
+        this.applicationsClient = checkNotNull(applicationsClient);
+        this.environment = checkNotNull(environment);
+    }
+
+    public static ApiKeyApplicationFetcher create(ApplicationsClient applicationsClient) {
+        return new ApiKeyApplicationFetcher(
+                applicationsClient,
+                Environment.parse(Configurer.getPlatform())
```
Dalamar42: This should be passed in via the constructor so configuration is given to this class instead of it searching for it itself.

atlas-api/src/main/java/org/atlasapi/application/ApiKeyApplicationFetcher.java
```diff
@@ -0,0 +1,69 @@
...
+        );
+    }
+
+    @Override
+    public ImmutableSet<String> getParameterNames() {
```
Dalamar42: Why do we need to expose this? The query parameter feels like an internal implementation detail of this class, not something that should be part of its public API.

atlas-api/src/main/java/org/atlasapi/application/ApiKeyApplicationFetcher.java
```diff
@@ -0,0 +1,69 @@
...
+        } catch (NullPointerException e) {
+            throw InvalidApiKeyException.create(e.getMessage(), "No api key supplied");
+        }
+
+        Result result = applicationsClient.resolve(
+                Query.create(apiKey, environment)
+        );
+
+        if (result.getErrorCode().isPresent()) {
+            throw InvalidApiKeyException.create(apiKey, result.getErrorCode().get().toString());
```
Dalamar42: It will be more obvious to something unfamiliar with the code if the exception message was something like:
`Applications service returned error code: <errorCode>`

atlas-api/src/main/java/org/atlasapi/application/ApplicationPersistenceModule.java
```diff
@@ -1,95 +1,25 @@
...
-        DatabasedMongo mongo = persistence.databasedWriteMongo();
-        IdGenerator idGenerator = new MongoSequentialIdGenerator(mongo, "application");
-        MongoApplicationStore legacyStore = new MongoApplicationStore(
-                mongo,
-                idGenerator,
-                ReadPreference.primaryPreferred()
+    public ApplicationsClient applicationsClient() {
+        return ApplicationsClientImpl.create(
+                Configurer.get("applications.client.host").get(),
+                new MetricRegistry()
```
Dalamar42: You need to pass the correct MetricRegistry here by autowiring the metricsModule

atlas-api/src/main/java/org/atlasapi/application/ApplicationPersistenceModule.java
```diff
@@ -1,95 +1,25 @@
...
-    public LegacyApplicationStore applicationStore() {
-        DatabasedMongo mongo = persistence.databasedWriteMongo();
-        IdGenerator idGenerator = new MongoSequentialIdGenerator(mongo, "application");
-        MongoApplicationStore legacyStore = new MongoApplicationStore(
-                mongo,
-                idGenerator,
-                ReadPreference.primaryPreferred()
+    public ApplicationsClient applicationsClient() {
+        return ApplicationsClientImpl.create(
+                Configurer.get("applications.client.host").get(),
```
Dalamar42: By convention we set configuration values to fields and then reference those fields so all the `Configurer.get(...` bits are concentrated at the top of the class

atlas-api/src/main/java/org/atlasapi/application/www/ApplicationWebModule.java
```diff
@@ -213,440 +98,15 @@ SelectionBuilder selectionBuilder() {
...
-        );
-    }
-
-    @Bean
-    protected EntityListWriter<User> usersListWriter() {
-        return new UsersListWriter(idCodec, sourceIdCodec);
-    }
-
     public
     @Bean
```
Dalamar42:
```
@Bean
public
```

atlas-api/src/main/java/org/atlasapi/application/www/ApplicationWebModule.java
```diff
@@ -213,440 +98,15 @@ SelectionBuilder selectionBuilder() {
 
     public
```
Dalamar42:
```
@Bean
public
```

atlas-api/src/main/java/org/atlasapi/output/writers/SourceWithIdWriter.java
```diff
@@ -48,4 +55,4 @@ public String listName() {
         return listName;
     }
 
-}
+}
```
Dalamar42: Newline

atlas-core/src/main/java/org/atlasapi/application/DefaultApplication.java
```diff
@@ -0,0 +1,68 @@
...
+        ApplicationConfiguration configuration = ApplicationConfiguration.builder()
+                .withPrecedence(finalPublishers)
+                .withEnabledWriteSources(ImmutableList.of())
+                .build();
+
+        return Application.builder()
+                .withId(-1L)
+                .withTitle("defaultApplication")
+                .withDescription("Default application")
+                .withEnvironment(null)
```
Dalamar42: Doesn't that blow up with an NPE?

atlas-core/src/test/java/org/atlasapi/output/OutputContentMergerTest.java
```diff
@@ -643,16 +563,21 @@ public void testMergesBroadcastsWithSimilarStartTimes() {
...
         assertThat(Iterables.getOnlyElement(merged.getBroadcasts()).getAliases().size(), is(2));
     }
 
     @Test
     public void mergeBroadcastsMergesTheirFields() {
-        ApplicationSources sources = sourcesWithPrecedence(false, Publisher.BBC, Publisher.PA)
-                .copy()
-                .withContentHierarchyPrecedence(ImmutableList.of(Publisher.BBC, Publisher.PA))
-                .build();
+//        ApplicationSources sources = sourcesWithPrecedence(false, Publisher.BBC, Publisher.PA)
```
Dalamar42: Commented out code

atlas-api/src/main/java/org/atlasapi/application/ApiKeyApplicationFetcher.java
jamiesprax: It's used in Query and Context parsers for building a list of required parameters

PR #510, opened 2017-01-31
Swallow exceptions for individual itemRefs when migrating content

emilssolmanis:
We used to have a bug that caused us to only migrate itemRefs and seriesRefs that were already present in C*. Now that we've fixed that, we're migrating all refs we get from Mongo. It turns out that some of those refs don't actually exist in Mongo, because hail Satan. This makes the code move on and try the rest of the refs instead of failing fast.


AdamDM:
lgtm

PR #516, opened 2017-02-09
Change CrewMember deserialisation to support null roles

Dalamar42:


atlas-cassandra/src/test/java/org/atlasapi/content/v2/serialization/CrewMemberSerializationTest.java
```diff
@@ -0,0 +1,38 @@
...
+    @Before
+    public void setUp() throws Exception {
+        serialization = new CrewMemberSerialization();
+    }
+
+    @Test
+    public void deserializationHandlesNullRoles() throws Exception {
+        CrewMember crewMember = new CrewMember("uri", "", Publisher.METABROADCAST);
+        //noinspection ConstantConditions
+        crewMember.withRole(null);
```
AdamDM: could put .withRole(null) on the line above as it returns itself like a builder, but totally optional

AdamDM:
lgtm

PR #525, opened 2017-02-22
Mbst 17721/broadcast merging

jamiesprax:
This returns a new block of broadcasts which may or may not be merged depending on the parent with the aim of returning the next most relevant broadcast first.

1- Filter out by channels in the platform
2- Filter out broadcasts in the past
3- Group by start time and channel parent
4- Construct representation for groups per start time/channel parent
5- Select key group (next broadcast) and apply downweigh

more detailed specs are in the docs


atlas-api/src/main/java/org/atlasapi/output/annotation/AggregatedBroadcastsAnnotation.java
```diff
@@ -0,0 +1,77 @@
...
+    private void writeAggregatedBroadcasts(
+            Item item,
+            FieldWriter writer,
+            OutputContext ctxt
+    ) throws IOException {
+        List<Id> downweighChannelIds = Arrays.stream(
+                ctxt.getRequest()
+                        .getParameter(Attributes.DOWNWEIGH.externalName())
+                        .split(","))
+                .map(Id::valueOf)
```
jamiesprax: This is wrong, needs decoding

atlas-api/src/main/java/org/atlasapi/output/annotation/AggregatedBroadcastsAnnotation.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    private void writeAggregatedBroadcasts(
+            Item item,
+            FieldWriter writer,
+            OutputContext ctxt
+    ) throws IOException {
+        List<Id> downweighChannelIds = Arrays.stream(
+                ctxt.getRequest()
+                        .getParameter(Attributes.DOWNWEIGH.externalName())
```
Dalamar42: If the parameter has not been provided will `.getParameter()` return `""` or `null`? We might NPE here

atlas-api/src/main/java/org/atlasapi/output/writers/AggregatedBroadcastWriter.java
```diff
@@ -0,0 +1,116 @@
...
+    private final NumberToShortStringCodec codec;
+    private final AliasWriter aliasWriter;
+    private final BroadcastIdAliasMapping aliasMapping;
+    private final BlackoutRestrictionWriter blackoutRestrictionWriter;
+    private final ChannelWriter channelWriter;
+
+    private AggregatedBroadcastWriter(NumberToShortStringCodec codec) {
+        this.codec = codec;
+
+        aliasWriter = new AliasWriter();
```
Dalamar42: For consistency I think it's nicer if we stick with the `this.` pattern for field assignments in constructors. Even if those don't technically need it

atlas-api/src/main/java/org/atlasapi/output/writers/AggregatedBroadcastWriter.java
```diff
@@ -0,0 +1,116 @@
...
+        aliasMapping = new BroadcastIdAliasMapping();
```
Dalamar42: Isn't it easier to test this class if we externalise these collaborators?

atlas-api/src/main/java/org/atlasapi/output/writers/AggregatedBroadcastWriter.java
```diff
@@ -0,0 +1,116 @@
...
+        Broadcast broadcast = entity.getBroadcast();
+        Channel channel = entity.getResolvedChannel().getChannel();
+
+        ImmutableList.Builder<Alias> aliases = ImmutableList.builder();
+        Alias idAlias = aliasMapping.apply(broadcast);
+        if (idAlias != null) {
+            aliases.add(idAlias);
+        }
+        aliases.addAll(broadcast.getAliases());
+        writer.writeList(aliasWriter, aliases.build(), ctxt);
```
Dalamar42: Why can't we use the existing `BroadcastWriter` here? As is we are duplicating the presentation logic

atlas-api/src/main/java/org/atlasapi/query/v4/channel/ChannelWriter.java
```diff
@@ -78,7 +94,21 @@ public void write(
         Channel channel = entity.getChannel();
 
         format.writeField("title", channel.getTitle());
-        format.writeField("id", idCode.encode(channel.getId().toBigInteger()));
+
+        if (contextHasAnnotation(ctxt, Annotation.AGGREGATED_BROADCASTS)) {
```
Dalamar42: I think this can introduce a bug. If we have both the `BROADCASTS` and `AGGREGATED_BROADCASTS` annotation enabled this will print the variants of the channels for both.

We can see if variants are in the entity and if so print them, otherwise skip that field.

atlas-api/src/main/java/org/atlasapi/query/QueryWebModule.java
```diff
@@ -518,6 +520,14 @@ private QueryAttributeParser contentQueryAttributeParser() {
...
                                 Attributes.SPECIALIZATION,
                                 EnumCoercer.create(Specialization.FROM_KEY())
+                        ),
+                        QueryAtomParser.create(
+                                Attributes.PLATFORM,
+                                IdCoercer.create(idCodec())
+                        ),
+                        QueryAtomParser.create(
+                                Attributes.DOWNWEIGH,
+                                IdCoercer.create(idCodec())
```
Dalamar42: I can't remember if we need to pass a flag to the coercer to tell it we expect this param might have multiple values passed to it instead of just a single one

atlas-api/src/main/java/org/atlasapi/query/v4/search/ContentQueryResultWriter.java
```diff
@@ -68,29 +70,68 @@ protected void writeResult(QueryResult<Content> result, ResponseWriter writer)
...
+
+        return OutputContext.valueOf(
+                queryContext,
+                resolveRegion(regionParam),
+                resolvePlatform(platformParam)
+        );
+    }
+
+    @Nullable
+    private Region resolveRegion(String id) throws IOException {
```
Dalamar42: `Optional<Region>` instead of null please

atlas-api/src/main/java/org/atlasapi/query/v4/search/ContentQueryResultWriter.java
```diff
@@ -68,29 +70,68 @@ protected void writeResult(QueryResult<Content> result, ResponseWriter writer)
...
+        ChannelGroup<?> channelGroup = resolveChannelGroup(id);
+        if (channelGroup instanceof Platform) {
+            return (Region) channelGroup;
+        } else {
+            Throwables.propagate(new NotAcceptableException(
```
Dalamar42: You don't need the `return null` here. You can do:
```
throw Throwables.propagate();
```

That said, the Guava peeps have deprecated `Throwables.propagate()` in Guava 20 and even though we are not using that yet we kind of started following their example and doing instead:
```
throw new RuntimeException(exception);
```

atlas-api/src/main/java/org/atlasapi/query/v4/search/ContentQueryResultWriter.java
```diff
@@ -68,29 +70,68 @@ protected void writeResult(QueryResult<Content> result, ResponseWriter writer)
...
-                                    channelGroup.get().getType()
-                            )
+                    String.format(
+                            "%s is a channel group of type '%s', should be 'platform",
+                            id,
+                            channelGroup
                     )
-            );
+            ));
+            return null;
```
Dalamar42: Ditto

atlas-api/src/main/java/org/atlasapi/query/v4/search/ContentQueryResultWriter.java
Dalamar42: That's irrelevant given the next comment actually

atlas-api/src/main/java/org/atlasapi/query/v4/search/ContentQueryResultWriter.java
```diff
@@ -68,29 +70,68 @@ protected void writeResult(QueryResult<Content> result, ResponseWriter writer)
...
+                1,
+                TimeUnit.MINUTES
+        ).getResources().first();
+
+        if (channelGroupOptional.isPresent()) {
+            return channelGroupOptional.get();
+        } else {
+           Throwables.propagate(new NotFoundException(id));
+        }
+        return null;
```
Dalamar42: Ditto

atlas-core/src/main/java/org/atlasapi/content/AggregatedBroadcast.java
```diff
@@ -0,0 +1,34 @@
...
+public class AggregatedBroadcast {
+
+    private final Broadcast broadcast;
+    private final ResolvedChannel resolvedChannel;
+    private final boolean channelIsParent;
+
+    private AggregatedBroadcast(Broadcast broadcast, ResolvedChannel resolvedChannel) {
+        this.broadcast = checkNotNull(broadcast);
+        this.resolvedChannel = checkNotNull(resolvedChannel);
+        channelIsParent = !broadcast.getChannelId().equals(resolvedChannel.getChannel().getId());
```
Dalamar42: Ditto `this.`

Dalamar42: This logic is not very obvious to someone unfamiliar with this code. Can we just pass the channel that should be used here always and remove the `channelIsParent` flag? This way the writer just uses whatever channel it's been given. It doesn't have to know what that channel is

atlas-core/src/main/java/org/atlasapi/content/AggregatedBroadcast.java
```diff
@@ -0,0 +1,34 @@
+package org.atlasapi.content;
+
+import org.atlasapi.channel.ResolvedChannel;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class AggregatedBroadcast {
+
+    private final Broadcast broadcast;
+    private final ResolvedChannel resolvedChannel;
```
Dalamar42: You can put the `Channel` here. I don't think you need the `ResolvedChannel` wrapper

atlas-core/src/main/java/org/atlasapi/content/BroadcastAggregator.java
```diff
@@ -0,0 +1,241 @@
...
+                .filter(aggregatedBroadcast -> ids.contains(aggregatedBroadcast.getBroadcast()
+                        .getChannelId()))
+                .findFirst();
+
+        if (!keyBroadcast.isPresent()) {
+            return ImmutableSet.copyOf(aggregatedBroadcasts);
+        }
+
+        return broadcastBuilder.add(keyBroadcast.get())
+                .addAll(aggregatedBroadcasts.stream()
```
Dalamar42: The broadcasts after the key one should be sorted by start time. I'm not sure if we have sorted them before this step.

atlas-core/src/main/java/org/atlasapi/content/AggregatedBroadcast.java
jamiesprax: I need it as a ResolvedChannel so I can add the variations and pass it to the ChannelWriter which only takes a ResolvedChannel so i'd have to wrap it at some point anyway.

atlas-api/src/main/java/org/atlasapi/query/QueryWebModule.java
jamiesprax: Attributes.DOWNWEIGH is created with `IdAttribute.list("downweigh", Channel.class);` so that should work right?

atlas-api/src/main/java/org/atlasapi/output/writers/AggregatedBroadcastWriter.java
jamiesprax: The existing `BroadcastWriter` resolves the channel midway through writing the broadcast meaning it will override my channel which, for aggregated broadcasts, would be the parent channel

PR #528, opened 2017-03-01
Change content bootstrap to update equiv schedule store graphs

Dalamar42:
- Change content bootstrap to use a message sending equiv graph store
  so that equivalence changes will generate messages for the equiv
  schedule store


AdamDM:
lgtm

PR #530, opened 2017-03-02
Fix/single aggregated broadcast bug

jamiesprax:
Fixed single broadcasts from being ignored.
Fixed NPE when trying to find parent on a broadcasts channel when it didn't have one.

### DIFF
--- a/atlas-core/src/main/java/org/atlasapi/content/BroadcastAggregator.java
+++ b/atlas-core/src/main/java/org/atlasapi/content/BroadcastAggregator.java
@@ -85,6 +85,8 @@ public class BroadcastAggregator {
                         dateTime,
                         aggregateBroadcastsInternal(sameTimeBroadcasts)
                 );
+            } else {
+                aggregatedBroadcasts.put(dateTime, Iterables.getOnlyElement(sameTimeBroadcasts));
             }
         }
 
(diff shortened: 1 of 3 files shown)


atlas-core/src/main/java/org/atlasapi/content/BroadcastAggregator.java
```diff
@@ -100,9 +103,15 @@ public static BroadcastAggregator create(ChannelResolver channelResolver) {
...
         Multimap<ChannelRef, ResolvedBroadcast> parentChannelMap = broadcasts.stream()
                 .collect(MoreCollectors.toImmutableListMultiMap(
-                        resolvedBroadcast -> resolvedBroadcast.getResolvedChannel()
-                                .getChannel()
-                                .getParent(),
+                        resolvedBroadcast -> MoreObjects.firstNonNull(
+                                resolvedBroadcast.getResolvedChannel()
+                                        .getChannel()
+                                        .getParent(),
+                                new ChannelRef(
```
Dalamar42: We should have a `Channel#toRef` method to convert a channel to its ref. It will be better to add that since it will be useful elsewhere.

atlas-core/src/test/java/org/atlasapi/content/BroadcastAggregatorTest.java
```diff
@@ -23,6 +23,7 @@
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
```
Dalamar42: Can you add a test for the case where no parent channel can be resolved as well

PR #527, opened 2017-02-28
Mbst 18177/index query parser hack

Dalamar42:

PR #531, opened 2017-03-06
Revert "Mbst 18177/index query parser hack"

jamiesprax:
Reverts atlasapi/atlas-deer#527


jamiesprax:
This release was causing issues with certain query parameters. 
Known ones are "title.beginning" and multiple values for "order_by", both of which worked prior to this commits release.

PR #533, opened 2017-03-08
Review model changes

jamiesprax:


AdamDM:
lgtm

atlas-api/src/main/java/org/atlasapi/output/writers/ReviewsWriter.java
```diff
@@ -38,9 +38,16 @@ public String fieldName(Review entity) {
     public void write(@Nonnull Review entity, @Nonnull FieldWriter writer, @Nonnull OutputContext ctxt)
             throws IOException {
 
-        String language = (null != entity.getLocale()) ? entity.getLocale().getLanguage() : null;
+        String language = (entity.getLocale() != null) ? entity.getLocale().getLanguage() : null;
+        String reviewType = (entity.getReviewType() != null) ? entity.getReviewType().toKey() : null;
+
         writer.writeField("language", language);
```
dragospitica: if the entity.getLocale() is null, will it not throw an NPE?

dragospitica: if `entity.getLocale()` is null, will it not throw an NPE?

atlas-api/src/main/java/org/atlasapi/output/writers/ReviewsWriter.java
```diff
@@ -38,9 +38,16 @@ public String fieldName(Review entity) {
...
         writer.writeField("review", entity.getReview());
+        writer.writeField("author", entity.getAuthor());
+        writer.writeField("author_initials", entity.getAuthorInitials());
+        writer.writeField("rating", entity.getRating());
+        writer.writeField("date", entity.getDate());
+        writer.writeField("review_type", reviewType);
```
dragospitica: ditto ^

atlas-api/src/main/java/org/atlasapi/output/writers/ReviewsWriter.java
jamiesprax: nope

atlas-api/src/main/java/org/atlasapi/output/writers/ReviewsWriter.java
jamiesprax: ditto nope

PR #536, opened 2017-03-14
Add filter to reject updates that would result in very large graphs

Dalamar42:
- These graphs are the result of bugs in equivalence and they cause
  significant problems in the downstream equivalent content and
  schedule stores. This prevents these graphs from being generated in
  Deer and also adds monitoring so we can have visibility when this
  happens

### DIFF
--- a/atlas-core/src/main/java/org/atlasapi/equivalence/EquivalenceGraphRejectionFilter.java
+++ b/atlas-core/src/main/java/org/atlasapi/equivalence/EquivalenceGraphRejectionFilter.java
@@ -1,0 +1,112 @@
+package org.atlasapi.equivalence;
+
+import org.atlasapi.entity.ResourceRef;
+
+import com.metabroadcast.common.stream.MoreCollectors;
+
+import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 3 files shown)


atlas-core/src/main/java/org/atlasapi/equivalence/EquivalenceGraphRejectionFilter.java
```diff
@@ -0,0 +1,108 @@
...
+            EquivalenceGraphUpdate graphUpdate
+    ) {
+        Decision decision = Decision.OK;
+
+        if (shouldWarn(subject, graphUpdate)) {
+            decision = Decision.WARN;
+        }
+
+        if (shouldReject(existingGraphs, graphUpdate)) {
+            decision = Decision.FAIL;
```
emilssolmanis: You can flip these and return early, no? I.e., default to `OK`, check for `FAIL`, then `WARN`.

Dalamar42: I want the logging in `shouldWarn` to happen even on a fail, but that might not be terribly obvious

PR #537, opened 2017-03-15
Fix missing publisher on some RT reviews

jamiesprax:
If content was removed from the feed, it will not have a source
as that field is now passed through from the ingester rather than
being populated in Deer. This defaults the publisher to match the
content publisher in the case where it is absent.

### DIFF
--- a/atlas-api/src/test/java/org/atlasapi/output/annotation/ReviewsAnnotationTest.java
+++ b/atlas-api/src/test/java/org/atlasapi/output/annotation/ReviewsAnnotationTest.java
@@ -1,0 +1,72 @@
+package org.atlasapi.output.annotation;
+
+import com.google.common.collect.ImmutableSet;
+import org.atlasapi.content.Content;
+import org.atlasapi.content.Film;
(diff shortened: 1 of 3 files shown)


AdamDM:
lgtm

atlas-api/src/test/java/org/atlasapi/output/annotation/ReviewsAnnotationTest.java
```diff
@@ -0,0 +1,72 @@
...
+        EntityListWriter<Review> reviewWriter = new ReviewsWriter(publisherWriter);
+        reviewsAnnotation = new ReviewsAnnotation(reviewWriter);
+
+        when(content.getReviews()).thenReturn(populateReviews());
+        when(content.getSource()).thenReturn(Publisher.RADIO_TIMES);
+
+    }
+
+    @Test
+    public void missingSourcesAreAddedToReviews() throws Exception {
```
Dalamar42: I'd rename this method to make it clear that missing sources are populated from the content source

PR #540, opened 2017-04-19
Add health check

jamiesprax:
Add health checks from common-health


atlas-api/src/main/java/org/atlasapi/system/HealthModule.java
```diff
@@ -53,4 +57,33 @@ public void addProbes() {
...
+                persistenceModule.persistenceModule().getSession()
+        );
+
+        Probe astyanaxProbe = org.atlasapi.system.health.probes.AstyanaxProbe.create(
+                "astyanax",
+                persistenceModule.persistenceModule().getContext()
+        );
+
+        return ImmutableList.of(
+                metricProbeFor(cassandraProbe),
```
Dalamar42: We also need a mongo probe. Ideally an ElasticSearch one as well, but I think we don't have one for that.

atlas-api/src/main/java/org/atlasapi/system/HealthModule.java
```diff
@@ -53,4 +57,33 @@ public void addProbes() {
...
+                metricProbeFor(astyanaxProbe)
+        );
+    }
+
+    private com.metabroadcast.common.health.probes.MetricsProbe metricProbeFor(
+            Probe probe
+    ) {
+        return com.metabroadcast.common.health.probes.MetricsProbe.builder()
```
Dalamar42: I thought health checks were changed to automatically report metrics when created.

atlas-api/src/main/java/org/atlasapi/system/health/HealthController.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @RequestMapping("/system/health/probes")
+    public void showHealthForProbes(HttpServletResponse response) throws IOException {
+        Result result = health.status(Health.FailurePolicy.ANY);
+        response.setStatus(result.getStatus() == Status.HEALTHY ? SC_OK
+                                                                : SC_INTERNAL_SERVER_ERROR
+        );
+
+        PrintWriter writer = response.getWriter();
+        result.getProbeResults().forEach(probeResult -> {
```
Dalamar42: Why not just serialise `ProbeResults` to JSON and print that? This seems more complicated than we need

atlas-api/src/main/java/org/atlasapi/system/health/probes/AstyanaxProbe.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    public Callable<ProbeResult> createRequest() {
+        return () -> {
+            try {
+                ConnectionPoolMonitor pool = context.getConnectionPoolMonitor();
+                List<String> infoList = Lists.newArrayList();
+
+                infoList.add("Socket timeouts: " + Long.toString(pool.getSocketTimeoutCount()));
```
Dalamar42: You don't need to preserve all this info. These were only around because before the health checks mixed and health and metric responsibilities which we shouldn't be doing anyway

PR #542, opened 2017-05-02
Add logging to debug bootstrap workers

jamiesprax:
There are a lot of class cast exceptions so gathering some more detailed
information will help locate the source of the issue.


AdamDM:
lgtm

PR #546, opened 2017-05-17
Merge Continuations in aggrecated broadcasts

TjlHope:
As continuations were showing up as seperate broadcasts.

### DIFF
--- a/atlas-core/src/main/java/org/atlasapi/content/Broadcast.java
+++ b/atlas-core/src/main/java/org/atlasapi/content/Broadcast.java
@@ -384,11 +384,18 @@ public class Broadcast extends Identified implements Hashable {
     }
 
     public Broadcast copy() {
+        // Use copyWithNewInterval with the current interval so we don't add code dupe
         return copyWithNewInterval(transmissionInterval);
     }
 
-    public Broadcast copyWithNewInterval(Interval transmissionInterval) {
-        Broadcast copy = new Broadcast(channelId, transmissionInterval);
+    /**
+     * Return a copy of this broadcast, but with a new transmission interval.
+     * Used for merging Broadcast continuations.
+     * @param newTransmissionInterval   the new transmission interval.
+     * @return  a new copy of this broadcast, with the new transmission interval.
+     */
+    public Broadcast copyWithNewInterval(Interval newTransmissionInterval) {
+        Broadcast copy = new Broadcast(channelId, newTransmissionInterval);
         Identified.copyTo(this, copy);
         copy.activelyPublished = activelyPublished;
         copy.sourceId = sourceId;
--- a/atlas-core/src/main/java/org/atlasapi/content/Broadcast.java
+++ b/atlas-core/src/main/java/org/atlasapi/content/Broadcast.java
@@ -388,31 +388,28 @@ public class Broadcast extends Identified implements Hashable {
         return copyWithNewInterval(transmissionInterval);
     }
 
-    /**
-     * Return a copy of this broadcast, but with a new transmission interval.
-     * Used for merging Broadcast continuations.
-     * @param newTransmissionInterval   the new transmission interval.
-     * @return  a new copy of this broadcast, with the new transmission interval.
-     */
     public Broadcast copyWithNewInterval(Interval newTransmissionInterval) {
         Broadcast copy = new Broadcast(channelId, newTransmissionInterval);
         Identified.copyTo(this, copy);
+        copy.scheduleDate = scheduleDate;
         copy.activelyPublished = activelyPublished;
         copy.sourceId = sourceId;
-        copy.scheduleDate = scheduleDate;
+        copy.versionId = versionId;
         copy.repeat = repeat;
         copy.subtitled = subtitled;
         copy.signed = signed;
         copy.audioDescribed = audioDescribed;
         copy.highDefinition = highDefinition;
         copy.widescreen = widescreen;
+        copy.surround = surround;
+        copy.live = live;
         copy.newSeries = newSeries;
         copy.newEpisode = newEpisode;
         copy.newOneOff = newOneOff;
         copy.premiere = premiere;
         copy.continuation = continuation;
-        copy.live = live;
-        copy.versionId = versionId;
+        copy.is3d = is3d;
+        copy.blackoutRestriction = blackoutRestriction;
         copy.revisedRepeat = revisedRepeat;
         return copy;
     }
--- a/atlas-core/src/main/java/org/atlasapi/content/Broadcast.java
+++ b/atlas-core/src/main/java/org/atlasapi/content/Broadcast.java
@@ -384,6 +384,10 @@ public class Broadcast extends Identified implements Hashable {
     }
 
     public Broadcast copy() {
+        return copyWithNewInterval(transmissionInterval);
+    }
+
+    public Broadcast copyWithNewInterval(Interval transmissionInterval) {
         Broadcast copy = new Broadcast(channelId, transmissionInterval);
         Identified.copyTo(this, copy);
         copy.activelyPublished = activelyPublished;
--- a/atlas-core/src/main/java/org/atlasapi/content/BroadcastAggregator.java
+++ b/atlas-core/src/main/java/org/atlasapi/content/BroadcastAggregator.java
@@ -2,7 +2,6 @@ package org.atlasapi.content;
 
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableMultimap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
@@ -22,6 +21,7 @@ import org.atlasapi.entity.Id;
 import org.atlasapi.entity.ResourceRef;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeComparator;
+import org.joda.time.Interval;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -29,10 +29,13 @@ import javax.annotation.Nullable;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.Map;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
@@ -61,6 +64,7 @@ public class BroadcastAggregator {
         if (platformOptional.isPresent()) {
             broadcasts = removeBroadcastsNotOnPlatform(broadcasts, platformOptional.get());
         }
+        broadcasts = aggregateContinuations(broadcasts);
 
         // Filter out previous broadcasts and collect by transmission time
         Multimap<DateTime, ResolvedBroadcast> broadcastMap = broadcasts.stream()
@@ -72,7 +76,6 @@ public class BroadcastAggregator {
                 .collect(MoreCollectors.toImmutableListMultiMap(
                         resolvedBroadcast -> resolvedBroadcast.getBroadcast().getTransmissionTime(),
                         resolvedBroadcast -> resolvedBroadcast
-
                 ));
 
         // Aggregate the broadcasts with same transmission times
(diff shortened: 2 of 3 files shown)


atlas-core/src/main/java/org/atlasapi/content/Broadcast.java
```diff
@@ -384,6 +384,10 @@ public BroadcastRef toRef() {
     }
 
     public Broadcast copy() {
+        return copyWithNewInterval(transmissionInterval);
```
AdamDM: why have a method simply referencing another method immediately, or is this important for backwards compatibility (things using the copy method that would otherwise need changing all over the place)?

atlas-core/src/main/java/org/atlasapi/content/BroadcastAggregator.java
```diff
@@ -61,6 +64,7 @@ public static BroadcastAggregator create(ChannelResolver channelResolver) {
         if (platformOptional.isPresent()) {
             broadcasts = removeBroadcastsNotOnPlatform(broadcasts, platformOptional.get());
         }
+        broadcasts = aggregateContinuations(broadcasts);
```
AdamDM: consider not reusing variables, if they've been processed somehow maybe use 'aggregateBroadcasts' instead of 'broadcasts'

atlas-core/src/test/java/org/atlasapi/content/BroadcastAggregatorTest.java
```diff
@@ -27,9 +27,7 @@
 import java.util.Set;
 
 import static org.hamcrest.core.Is.is;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertThat;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
```
AdamDM: star imports are against our style guide for various reasons, check out our style guide and have a look in the engineering repo for the intellij configuration

AdamDM:
otherwise looks great!

atlas-core/src/test/java/org/atlasapi/content/BroadcastAggregatorTest.java
```diff
@@ -105,6 +103,42 @@ public void broadcastOnSameChannelAtDifferentTimesDoNotAggregate() throws Except
...
+        assertThat(resolvedBroadcasts.size(), is(1));
+        Broadcast broadcast = Iterables.getFirst(resolvedBroadcasts, null).getBroadcast();
+        assertEquals(firstBroadcast.getTransmissionTime(), broadcast.getTransmissionTime());
+        assertEquals(secondBroadcast.getTransmissionEndTime(), broadcast.getTransmissionEndTime());
+    }
+
+    @Test
     public void downweighChannelBroadcastIsNotFirst() throws Exception {
 
         Broadcast downweighedBroadcast1 = new Broadcast(Id.valueOf(333L), DateTime.now().plusHours(1), DateTime.now().plusHours(2));
```
jamiesprax: When going through existing code its quite nice to do a bit of cleaning up of things around it. This, and the following assignments can use the method `getFutureBroadcast(333L, 1, 2)` which someone _(me)_ missed when writing it.

atlas-core/src/main/java/org/atlasapi/content/BroadcastAggregator.java
```diff
@@ -251,6 +254,21 @@ private ResolvedChannel buildResolvedChannel(
...
 
+    Set<Broadcast> aggregateContinuations(Set<Broadcast> broadcasts) {
+        Map<Id, Broadcast> channelId2Broadcast = new HashMap<>();
+        Set<Broadcast> aggregated = new LinkedHashSet<>(broadcasts.size());
+        for (Broadcast broadcast : broadcasts) {
+            Broadcast old = channelId2Broadcast.put(broadcast.getChannelId(), broadcast);
+            if (old != null && Boolean.TRUE.equals(broadcast.getContinuation())) {
+                broadcast = old.copyWithNewInterval(
+                        new Interval(old.getTransmissionTime(), broadcast.getTransmissionEndTime()));
+                aggregated.remove(old);
```
jamiesprax: If this is the case then `aggregated` can be an ImmutableSet.Builder and then this method returns `.build()` at the end

atlas-core/src/main/java/org/atlasapi/content/BroadcastAggregator.java
```diff
@@ -251,6 +254,21 @@ private ResolvedChannel buildResolvedChannel(
                 .collect(MoreCollectors.toImmutableSet());
     }
 
+    Set<Broadcast> aggregateContinuations(Set<Broadcast> broadcasts) {
+        Map<Id, Broadcast> channelId2Broadcast = new HashMap<>();
+        Set<Broadcast> aggregated = new LinkedHashSet<>(broadcasts.size());
```
dragospitica: for consistency use guava collections: `Maps.newHashMap()`, `Sets.newLinkedHashSet()` or `Sets.newLinkedHashSetWithExpectedSize()`

atlas-core/src/main/java/org/atlasapi/content/Broadcast.java
TjlHope: Yes, it's for backwards compat.

atlas-core/src/main/java/org/atlasapi/content/BroadcastAggregator.java
```diff
@@ -251,6 +267,48 @@ private ResolvedChannel buildResolvedChannel(
...
+                                    original.getTransmissionTime(),
+                                    broadcast.getTransmissionEndTime())));
+        }
+    }
+    private ListMultimap<Id, Broadcast> combineBroadcastContinuations(
+            ListMultimap<Id, Broadcast> broadcasts1,
+            ListMultimap<Id, Broadcast> broadcasts2
+    ) {
+        broadcasts2.values().forEach(b -> accumulateBroadcastContinuations(broadcasts1, b));
+        return broadcasts1;
```
AdamDM: I don't particularly like altering a variable using a static method, it won't actually work unless it's a field and I can't see that. Re my earlier comment consider making the accumulateBroadcastContinuations method return something.

atlas-core/src/main/java/org/atlasapi/content/Broadcast.java
AdamDM: consider putting the contents of copyWithNewInterval into the copy method if all it's going to do is call that

AdamDM: Ignore me, Jamie has explained it's purpose. Note that it's not obvious at first glance! perhaps add a comment to the copy() method that states its going to copy with its original interval (not a new one) to differentiate the methods

atlas-core/src/main/java/org/atlasapi/content/BroadcastAggregator.java
```diff
@@ -251,6 +267,48 @@ private ResolvedChannel buildResolvedChannel(
...
+                .sorted(Comparator.comparing(Broadcast::getTransmissionEndTime))
+                .collect(
+                        Collector.<Broadcast, ListMultimap<Id, Broadcast>, Set<Broadcast>>of(
+                                ArrayListMultimap::create,
+                                this::accumulateBroadcastContinuations,
+                                this::combineBroadcastContinuations,
+                                m -> ImmutableSet.copyOf(m.values())));
+    }
+
+    private void accumulateBroadcastContinuations(
```
AdamDM: OK I see how this method works re discussion in chat

atlas-core/src/test/java/org/atlasapi/content/BroadcastAggregatorTest.java
```diff
@@ -105,11 +107,93 @@ public void broadcastOnSameChannelAtDifferentTimesDoNotAggregate() throws Except
...
+        );
+
+        assertThat(resolvedBroadcasts.size(), is(1));
+        Broadcast broadcast = Iterables.getFirst(resolvedBroadcasts, null).getBroadcast();
+        assertThat(broadcast.getTransmissionTime(), is(firstBroadcast.getTransmissionTime()));
+        assertThat(broadcast.getTransmissionEndTime(), is(secondBroadcast.getTransmissionEndTime()));
+    }
+
+    @Test
+    public void broadcastContinuationOnSameChannelOutOfOrder() throws Exception {
```
jamiesprax: This test looks identical to the one at ln:110

atlas-core/src/main/java/org/atlasapi/content/BroadcastAggregator.java
TjlHope: The return is actually right at the top of the method. All the confusion is really making me think this would have been better without the stream stuff...

atlas-core/src/main/java/org/atlasapi/content/BroadcastAggregator.java
```diff
@@ -57,13 +71,17 @@ public static BroadcastAggregator create(ChannelResolver channelResolver) {
             Optional<Platform> platformOptional,
             List<Id> downweighChannelIds
     ) {
+
         // Remove broadcasts we don't care about
-        if (platformOptional.isPresent()) {
-            broadcasts = removeBroadcastsNotOnPlatform(broadcasts, platformOptional.get());
-        }
+        Set<Broadcast> platformBroadcasts = platformOptional.isPresent()
```
TjlHope: Don't think that'll help in this case, as it seems to only want a consumer (i.e. no return value), but I'll keep that in mind for the future.

atlas-core/src/main/java/org/atlasapi/content/BroadcastAggregator.java
TjlHope: Unfortunately this seems to be how Collectors have to work, agree it's not very clean...

atlas-core/src/test/java/org/atlasapi/content/BroadcastAggregatorTest.java
TjlHope: There's an ordering difference, as I noticed it would break the collector if they were out of order

TjlHope:
Just for comparison, here's a version using an inner class to contain it all a bit more:
```java
    Set<Broadcast> mergeBroadcastContinuations(Iterable<Broadcast> broadcasts) {
        return StreamSupport.stream(broadcasts.spliterator(), false)
                .sorted(Comparator.comparing(Broadcast::getTransmissionEndTime))
                .collect(BroadcastContinuationCollector.instance());
    }

    static class BroadcastContinuationCollector
    implements Collector<Broadcast, ListMultimap<Id, Broadcast>, Set<Broadcast>> {
        private static final BroadcastContinuationCollector INSTANCE =
                new BroadcastContinuationCollector();
        public static BroadcastContinuationCollector instance() { return INSTANCE; }

        @Override public Supplier<ListMultimap<Id, Broadcast>> supplier() {
            return ArrayListMultimap::create;
        }

        @Override public BiConsumer<ListMultimap<Id, Broadcast>, Broadcast> accumulator() {
            return (channelId2Broadcast, broadcast) -> {
                List<Broadcast> broadcasts = channelId2Broadcast.get(broadcast.getChannelId());
                if (broadcasts.isEmpty() || !Boolean.TRUE.equals(broadcast.getContinuation())) {
                    // If we don't have a broadcast on the channel yet, or if it's not a continuation,
                    // just add it.
                    broadcasts.add(broadcast);
                } else {
                    // Otherwise remove the previous broadcast, merge it with this one, and add the result.
                    int last = broadcasts.size() - 1;
                    Broadcast original = broadcasts.get(last);
                    broadcasts.set(
                            last,
                            original.copyWithNewInterval(
                                    new Interval(
                                            original.getTransmissionTime(),
                                            broadcast.getTransmissionEndTime())));
                }
            };
        }

        @Override public BinaryOperator<ListMultimap<Id, Broadcast>> combiner() {
            final BiConsumer<ListMultimap<Id, Broadcast>, Broadcast> accumulator = accumulator();
            return (broadcasts1, broadcasts2) -> {
                broadcasts2.values().forEach(b -> accumulator.accept(broadcasts1, b));
                return broadcasts1;
            };
        }

        @Override public Function<ListMultimap<Id, Broadcast>, Set<Broadcast>> finisher() {
            return broadcasts -> ImmutableSet.copyOf(broadcasts.values());
        }

        @Override public Set<Characteristics> characteristics() {
            return Collections.emptySet();
        }
    }
```

atlas-core/src/main/java/org/atlasapi/content/BroadcastAggregator.java
```diff
@@ -251,6 +267,61 @@ private ResolvedChannel buildResolvedChannel(
...
+     * resolving the continuations.
+     * @param channelId2Broadcast   accumulated collection of broadcasts.
+     * @param broadcast             the broadcast to fold into the collection.
+     */
+    private void accumulateBroadcastContinuations(
+            ListMultimap<Id, Broadcast> channelId2Broadcast,
+            Broadcast broadcast
+    ) {
+        List<Broadcast> broadcasts = channelId2Broadcast.get(broadcast.getChannelId());
+        if (broadcasts.isEmpty() || !Boolean.TRUE.equals(broadcast.getContinuation())) {
```
dragospitica: for the second condition you could do just `!broadcast.getContinuation()` since the method already returns a boolean value

TjlHope: It actually returns a Boolean - had it as null in a unit test which cased an NPE

PR #547, opened 2017-05-22
implement channel equivalence support

jamiesprax:


AdamDM:
lgtm

atlas-legacy/src/main/java/org/atlasapi/system/legacy/LegacyChannelTransformer.java
```diff
@@ -19,6 +26,8 @@
 public class LegacyChannelTransformer
         extends BaseLegacyResourceTransformer<org.atlasapi.media.channel.Channel, Channel> {
 
+    private final SubstitutionTableNumberCodec codec = SubstitutionTableNumberCodec.lowerCaseOnly();
+
     @Nullable
```
TjlHope: how can this be null?

PR #548, opened 2017-06-15
implement output merging when precedence is enabled

jamiesprax:


atlas-api/src/test/java/org/atlasapi/query/v4/channel/ChannelQueryExecutorTest.java
```diff
@@ -65,6 +65,14 @@
     @InjectMocks
     private ChannelQueryExecutor objectUnderTest;
 
+    @Mock
+    private Application application = mock(Application.class);
```
TjlHope: I don't think you need both the @Mock and the =mock(...)

atlas-core/src/main/java/org/atlasapi/channel/ResolvedChannel.java
```diff
@@ -13,18 +13,20 @@
...
             Optional<List<ChannelGroupSummary>> channelGroupSummaries,
             Optional<Channel> parentChannel,
             Optional<Iterable<Channel>> channelVariations,
             Optional<ChannelGroupMembership> channelGroupMembership,
-            java.util.Optional<List<ChannelVariantRef>> includedVariants,
-            java.util.Optional<List<ChannelVariantRef>> excludedVariants
+            Optional<List<ChannelVariantRef>> includedVariants,
+            Optional<List<ChannelVariantRef>> excludedVariants,
+            Optional<Iterable<Channel>> equivalents
     ) {
```
TjlHope: Don't know if it matters, but despite being 'final' all these collections are still potentially mutable as they're passed in directly from the builder, and they might have been given to the builder as mutables.
If we care we should probably be doing a defensive copy into an ImmutableList etc...

atlas-api/src/main/java/org/atlasapi/query/v4/channelgroup/ChannelGroupQueryExecutor.java
```diff
@@ -320,7 +320,7 @@ private ResolvedChannelGroup resolveAnnotationData(
     @Nullable
     private Iterable<Channel> resolveChannelEquivalents(Channel channel) {
 
-        if (channel.getSameAs().isEmpty()) {
+        if (channel.getSameAs() == null || channel.getSameAs().isEmpty()) {
```
TjlHope: You can potentially use CollectionUtils.isEmpty
But don't know if apache commons is in the path.
It would also be nicer to have channel.getSameAs() return an empty list instead of null, but I'm guessing that's too big a change?

PR #550, opened 2017-07-06
Container output merging

jamiesprax:


atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -88,6 +90,12 @@
     private static final Function<Described, String> TO_SHORT_DESCRIPTION =
             input -> input == null ? null : input.getShortDescription();
 
+    private static final Function<Item, ContainerRef> TO_CONTAINER_REF =
```
AdamDM: are these reused? Why not leave them inline?

PR #556, opened 2017-08-02
add param to force cache refresh

jamiesprax:


atlas-core/src/main/java/org/atlasapi/entity/IdResolver.java
```diff
@@ -8,4 +8,8 @@
 
     ListenableFuture<Resolved<I>> resolveIds(Iterable<Id> ids);
 
+    default ListenableFuture<Resolved<I>> resolveIds(Iterable<Id> ids, Boolean refreshCache) {
+        throw new UnsupportedOperationException();
```
TjlHope: should this just proxy to the plain resolveIds method?

jamiesprax: already did ;)

PR #557, opened 2017-08-08
Fix images filtering

TjlHope:
Filter the list of images even if there is not top match.

Also fix the bootstrap hierarchy to work with non-episode items.

### DIFF
--- a/atlas-core/src/test/java/org/atlasapi/output/OutputContentMergerTest.java
+++ b/atlas-core/src/test/java/org/atlasapi/output/OutputContentMergerTest.java
@@ -50,6 +50,7 @@ import org.joda.time.Interval;
 import org.junit.Test;
 
 import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThat;
 import static org.mockito.Mockito.mock;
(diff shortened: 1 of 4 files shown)


atlas-core/src/test/java/org/atlasapi/output/OutputContentMergerTest.java
```diff
@@ -473,24 +474,124 @@ public void testImageWithMerging() {
...
+
+        item2.setImage(image2.getCanonicalUri());
+        item2.setImages(ImmutableSet.of(image2));
+
+        Content merged = merger.merge(item2,  ImmutableList.of(item1), application);
+        assertNull(merged.getImage());
+        assertThat(merged.getImages().size(), is(0));
+    }
+
+    /* TODO @Test
```
AdamDM: why leave in a massive commented out test? If you put a todo in the codebase then add a jira ticket link

AdamDM:
olgtm

atlas-core/src/test/java/org/atlasapi/output/OutputContentMergerTest.java
TjlHope: Good spot - that was to remind me to remove it :-)

PR #558, opened 2017-08-21
remove timeshifted and HD from aggregated broadcasts

jamiesprax:


atlas-core/src/main/java/org/atlasapi/content/BroadcastAggregator.java
```diff
@@ -370,4 +373,9 @@ private boolean variantIdIsNotDownweighed(List<Id> downweighedIds, ResolvedBroad
         return downweighedIds.stream().noneMatch(includedVariantIds::contains);
     }
 
+    private boolean isNotTimeshiftedOrHd(Channel channel) {
+        return !(channel.getTimeshifted() ||
```
TjlHope: Can't this NPE as they're `Boolean`s (i.e. objects not primitives)

atlas-core/src/main/java/org/atlasapi/content/BroadcastAggregator.java
```diff
@@ -374,8 +374,10 @@ private boolean variantIdIsNotDownweighed(List<Id> downweighedIds, ResolvedBroad
     }
 
     private boolean isNotTimeshiftedOrHd(Channel channel) {
-        return !(channel.getTimeshifted() ||
-                channel.getHighDefinition());
+        boolean timeshifted = MoreObjects.firstNonNull(channel.getTimeshifted(), false);
```
TjlHope: You can use Boolean.TRUE.equals(obj) if you want, though it's less explicit, so I don't mind :-)

PR #559, opened 2017-08-22
Created bean for EventBootstrapController

dragospitica:


atlas-processing/src/main/java/org/atlasapi/system/bootstrap/BootstrapModule.java
```diff
@@ -135,6 +135,11 @@ IndividualTopicBootstrapController topicBootstrapController() {
     }
 
     @Bean
+    EventBootstrapController eventBootstrapController() {
```
jamiesprax: Surely it had a bean already otherwise the autowiring would have failed, no?

dragospitica: @jamiesprax I created the bean now, it wasn't there before

PR #563, opened 2017-10-30
update channel group filtering

jamiesprax:
Filter out channel groups on the channel that the application does not have read access to


atlas-core/src/main/java/org/atlasapi/channel/Channel.java
```diff
@@ -383,6 +383,11 @@ public Builder withChannelGroups(Iterable<ChannelGroupMembership> channelGroups)
             return this;
         }
 
+        public Builder withReplacementChannelGroups(Iterable<ChannelGroupMembership> channelGroups) {
```
TjlHope: Why `withReplacement`? Normally wouldn't the builder just have the field name?

jamiesprax: If you use the builder.fromChannel(channel) to start it off, it populates the channelGroups with the existing channel groups on the channel. The existing withChannelGroups methods adds the given memberships to the collection whereas this one replaces the channel groups with a new set

PR #566, opened 2017-11-27
Filter item summaries

jamiesprax:
Trying to serialize a brand that contains an ItemSummary without a title causes throws
an exception which interrupts the update of the Brand. This change is intended to log
the offending ItemSummary ID and allow the process to continue without it.

### DIFF
--- a/atlas-cassandra/src/main/java/org/atlasapi/content/ContentSerializationVisitor.java
+++ b/atlas-cassandra/src/main/java/org/atlasapi/content/ContentSerializationVisitor.java
@@ -6,8 +6,6 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import com.google.common.base.Strings;
-import com.google.common.collect.ImmutableList;
 import org.atlasapi.entity.Alias;
 import org.atlasapi.entity.Award;
 import org.atlasapi.entity.AwardSerializer;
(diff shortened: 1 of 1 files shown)


atlas-cassandra/src/main/java/org/atlasapi/content/ContentSerializationVisitor.java
```diff
@@ -291,13 +297,22 @@ private Builder visitContainer(Container container) {
...
-        );
+        
+        if (container.getItemSummaries() != null) {
+            builder.addAllItemSummaries(
+                    container.getItemSummaries().stream()
+                            .map(is -> {
+                                try {
+                                    return itemSummarySerializer.serialize(is).build();
+                                } catch (Exception e) {
+                                    log.error("ItemSummary without title: {}", is.getItemRef().getId(), e);
```
alianos-: How do we know the exception is without title? I think the exception itself should contain the information on what is missing. Also maybe catch the specific exception that is thrown?

atlas-cassandra/src/main/java/org/atlasapi/content/ContentSerializationVisitor.java
```diff
@@ -291,13 +297,22 @@ private Builder visitContainer(Container container) {
...
-                Iterables.transform(
-                        container.getItemSummaries(),
-                        is -> itemSummarySerializer.serialize(is).build()
-                )
-        );
+        
+        if (container.getItemSummaries() != null) {
+            builder.addAllItemSummaries(
+                    container.getItemSummaries().stream()
+                            .map(is -> {
```
alianos-: Can we break this down to at least a local var assignment? Its already messy having a try catch within the mapping function.

PR #570, opened 2017-12-22
Added support to migrate list of content IDs

dragospitica:


atlas-processing/src/main/java/org/atlasapi/system/debug/ContentDebugController.java
```diff
@@ -351,28 +354,34 @@ public void indexContent(
         }
     }
 
-    @RequestMapping("/system/debug/content/{id}/migrate")
+    @RequestMapping("/system/debug/content/migrate")
```
alianos-: Its probably safer to not change the endpoint, just create a new one. (that been said make sure it doesn't already exist because it has happened in the past)

atlas-processing/src/main/java/org/atlasapi/system/debug/ContentDebugController.java
```diff
@@ -351,28 +354,34 @@ public void indexContent(
...
+        if (Strings.isNullOrEmpty(ids)) {
+            throw new IllegalArgumentException("Must specify at least one content ID to migrate");
+        }
+        Iterable<String> requestedIds = commaSplitter.split(ids);
+        for (String id : requestedIds) {
+            try {
+                Long decodedId = lowercase.decode(id).longValue();
 
-            Content content = resolveLegacyContent(decodedId);
+                Content content = resolveLegacyContent(decodedId);
```
alianos-: and of course in the loop just call the method of the other endpoint (or move all functional code to a new method).

PR #575, opened 2018-04-16
Tjlh/type output merging

TjlHope:
Had to add all the non-static copyTo methods in the hierarchy to avoid needing reflection to get the proper types.

### DIFF
--- a/atlas-core/src/main/java/org/atlasapi/content/Content.java
+++ b/atlas-core/src/main/java/org/atlasapi/content/Content.java
@@ -53,7 +53,7 @@ public abstract class Content extends Described
     private Set<String> languages = ImmutableSet.of();
     private Set<Certificate> certificates = ImmutableSet.of();
     private Integer year = null;
-    private Set<Encoding> manifestedAs = Sets.newLinkedHashSet();
+    @Nullable private Set<Encoding> manifestedAs = Sets.newLinkedHashSet();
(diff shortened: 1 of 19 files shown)


atlas-core/src/main/java/org/atlasapi/content/Content.java
```diff
@@ -169,20 +169,29 @@ public Boolean isGenericDescription() {
...
         to.relatedLinks = from.relatedLinks;
         to.tags = from.tags;
         to.readHash = from.readHash;
-        to.people = Lists.newArrayList(Iterables.transform(from.people, CrewMember.COPY));
+        to.people = Lists.newArrayList(Iterables.transform(from.people, CrewMember::copy));
         to.languages = from.languages;
         to.certificates = from.certificates;
         to.year = from.year;
-        to.manifestedAs = Sets.newHashSet(from.manifestedAs);
+        to.manifestedAs = from.manifestedAs == null ? null : Sets.newLinkedHashSet(from.manifestedAs);
```
jamiesprax: Did you consider an empty set instead of null?

TjlHope: I guess I was trying to keep existing behaviour. I can quite easily change the class to never have a null manifestedAs, but wasn't sure if that would inadvertently break stuff.
Or is general principle to change stuff like this and fix what breaks later?

PR #577, opened 2018-04-24
Schedule bootstrap locking

tristanbeardmore-gray:


atlas-core/src/main/java/org/atlasapi/locks/ScheduleBootstrapLock.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    public void lock(Channel channel, Interval interval) {
+        for(Key k : getKeys(channel, interval)) {
+            getLock(k).lock();
+        }
+    }
+
+    public boolean tryLock(Channel channel, Interval interval) {
+        SortedSet<Key> keys = getKeys(channel, interval);
```
TjlHope: If you make this a NavigableSet, then you can do set.headSet(key) to get everything before that without having to do the equals check

atlas-core/src/main/java/org/atlasapi/locks/ScheduleBootstrapLock.java
```diff
@@ -0,0 +1,99 @@
...
+        for(Key key : keys) {
+            if(!getLock(key).tryLock()) {
+                for(Key k : keys) {
+                    if(k.equals(key)) {
```
TjlHope: It really feels like this should be a separate helper method

atlas-core/src/main/java/org/atlasapi/locks/ScheduleBootstrapLock.java
```diff
@@ -0,0 +1,99 @@
...
+                        break;
+                    }
+                    getLock(k).unlock();
```
TjlHope: Can unlock ever throw? - if so you need to try-catch all of these

atlas-core/src/main/java/org/atlasapi/locks/ScheduleBootstrapLock.java
```diff
@@ -0,0 +1,99 @@
...
+            getLock(k).unlock();
+        }
+    }
+
+    //All days in the interval
+    private SortedSet<Key> getKeys(Channel channel, Interval interval) {
+        String channelId = channel.getId().toString();
+        LocalDate date = interval.getStart().toLocalDate();
+        Set<Key> keys = new HashSet<>();
+        while(date.compareTo(interval.getEnd().toLocalDate()) <= 0) {
```
TjlHope: Probably better like:
```
LocalDate startDate = interval.getStart().toLocalDate();
LocalDate endDate = interval.getEnd().toLocalDate();
for (LocalDate date = startDate; !startDate.isAfter(endDate); date = date.plusDays(1)) {
```

atlas-core/src/main/java/org/atlasapi/locks/ScheduleBootstrapLock.java
```diff
@@ -0,0 +1,99 @@
...
+        for(Key k : getKeys(channel, interval)) {
+            getLock(k).unlock();
+        }
+    }
+
+    //All days in the interval
+    private SortedSet<Key> getKeys(Channel channel, Interval interval) {
+        String channelId = channel.getId().toString();
+        LocalDate date = interval.getStart().toLocalDate();
+        Set<Key> keys = new HashSet<>();
```
TjlHope: Can you not just use an ImmutableSortedSet.Builder directly?

atlas-core/src/main/java/org/atlasapi/locks/ScheduleBootstrapLock.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hashCode(channelId, date);
+        }
+
+        @Override
+        public int compareTo(Key that) {
+            return this.date.compareTo(that.date);
```
TjlHope: You need to include the channelId in here as well to not break stuff, e.g.
```
int comp = this.date.compareTo(that.date);
if (comp != 0) return comp;
comp = this.channelId.compareTo(that.channelId);
if (comp != 0) return comp;
return 0;
```

atlas-core/src/main/java/org/atlasapi/locks/ScheduleBootstrapLock.java
```diff
@@ -0,0 +1,99 @@
...
+        return ImmutableSortedSet.copyOf(keys);
+    }
+
+
+    private class Key implements Comparable<Key> {
+        private final String channelId;
+        private final LocalDate date;
+
+        Key(String channelId, LocalDate date) {
+            this.channelId = channelId;
```
TjlHope: `checkNotNull`s on these?

atlas-core/src/main/java/org/atlasapi/locks/ScheduleBootstrapLock.java
```diff
@@ -0,0 +1,99 @@
...
+
+public class ScheduleBootstrapLock {
+    private ConcurrentHashMap<Key, ReentrantLock> lockMap = new ConcurrentHashMap<>();
+
+    private ReentrantLock getLock(Key key) {
+        return lockMap.computeIfAbsent(key, (k) -> new ReentrantLock());
+    }
+
+    public void lock(Channel channel, Interval interval) {
+        for(Key k : getKeys(channel, interval)) {
```
TjlHope: put a comment here that it's only safe because the keys are from a sortedset, so it locks in a guaranteed order.

atlas-processing/src/main/java/org/atlasapi/system/bootstrap/ScheduleBootstrapController.java
```diff
@@ -233,15 +314,37 @@ private Interval interval(LocalDate day) {
...
+        }
+
+        public ScheduleBootstrapResponse(Collection<ScheduleBootstrapper.Status> status) {
+            this.status = ImmutableList.copyOf(checkNotNull(status));
+        }
+
+        public Collection<ScheduleBootstrapper.Status> getStatus() {
+            return status;
+        }
+
```
TjlHope: Should this have a
```
public boolean isBootstrapping() { return !status.isEmpty(); }
```
type method so it's more similar to the existing?

atlas-processing/src/main/java/org/atlasapi/system/bootstrap/ScheduleBootstrapper.java
```diff
@@ -51,53 +51,44 @@ public ScheduleBootstrapper(
...
     ) {
-        if (!bootstrapLock.tryLock()) {
-            return false;
-        }
-        bootstrapping.set(true);
-        processed.set(0);
-        failures.set(0);
-        progress.set(0);
-        total.set(Iterables.size(channels));
+        Status status = new Status(StreamUtils.stream(channels).map(c -> c.getId().toString()).collect(MoreCollectors.toImmutableList()),
```
TjlHope: `c.getId().toString()` doesn't seem a very safe way of doing it? Is there not a better way to get the string ID?
If the id is the long value, would you be better just using that directly in the schedulebootstaplock? Or use the encode method?
Also, this could be `.map(Channel::getId).map(Id::toString)` or chosen better impl.

atlas-processing/src/main/java/org/atlasapi/system/bootstrap/ScheduleBootstrapper.java
```diff
@@ -129,10 +133,10 @@ public void onSuccess(UpdateProgress result) {
                         channel.getId(),
                         channel.getTitle(),
                         result,
-                        progress.incrementAndGet(),
-                        total.get(),
-                        processed.incrementAndGet(),
-                        failures.get()
+                        status.progress.incrementAndGet(),
+                        status.total.get(),
```
TjlHope: I'm assuming total is set somewhere else then? Would be good to have to comments as to how these stats are linked together

PR #578, opened 2018-04-30
add annotation for including past broadcasts

jamiesprax:


atlas-core/src/main/java/org/atlasapi/content/BroadcastAggregator.java
```diff
@@ -72,8 +73,13 @@ public static BroadcastAggregator create(ChannelResolver channelResolver) {
         Set<Broadcast> mergedContinuations = mergeBroadcastContinuations(platformBroadcasts);
 
         // Filter out previous broadcasts and collect by transmission time
+        if (!includePastBroadcasts) {
+            mergedContinuations = mergedContinuations.stream()
+                    .filter(broadcast -> broadcast.getTransmissionEndTime().isAfterNow())
+                    .collect(Collectors.toSet());
+        }
+
```
TjlHope: You could keep a ref to the stream, rather than re-collecting to a set again.

PR #582, opened 2018-06-11
improve large equiv set handling

jamiesprax:
parallelize large set translation and reduce immutable set copying cost


tristanbeardmore-gray:
Merged these with my protobuf upgrade branch

PR #604, opened 2019-05-07
Add all broadcasts annotation

EmiEnache:
Add annotations with more relaxed filters. Purpose is to fix issue where we only show broadcasts from highest-precedence source in cases where it has only very old broadcasts, while there are lower-precedence sources with more recent (including upcoming) broadcasts.

https://metabroadcast.atlassian.net/browse/ENG-302

### DIFF
--- a/atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
+++ b/atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
@@ -9,7 +9,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
 import javax.annotation.Nullable;
@@ -628,7 +627,7 @@ public class OutputContentMerger implements EquivalentsMergeStrategy<Content> {
                 );
 
         //TODO write more comments on what/why (here and/or in JIRA ticket)
-        if(activeAnnotations.contains(Annotation.ALL_BROADCASTS)){
+        if (activeAnnotations.contains(Annotation.ALL_BROADCASTS)) {
             chosen.setBroadcasts(
                     all.stream()
                             .map(T::getBroadcasts)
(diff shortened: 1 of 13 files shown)


atlas-core/src/test/java/org/atlasapi/output/BroadcastMergingTest.java
```diff
@@ -47,7 +48,9 @@ public void testBroadcastMergingNoBroadcasts() {
         chosenItem.addEquivalentTo(notChosenItem);
         notChosenItem.addEquivalentTo(chosenItem);
 
-        executor.merge(chosenItem, ImmutableList.of(notChosenItem), application);
+        executor.merge(chosenItem, ImmutableList.of(notChosenItem), application,
+                Collections.emptySet()
```
dragospitica: There are no tests for the new annotations ntz ntz ntz

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -594,40 +602,53 @@ private void mergeFilmProperties(Application application, Film chosen,
...
                 .onResultOf(Sourceds.toPublisher())
-                .leastOf(StreamSupport.stream(all.spliterator(), false)
+                .leastOf(
+                        all.stream()
                                 .filter(HAS_BROADCASTS::apply)
                                 .collect(Collectors.toList()),
                         1
                 );
 
+        if (activeAnnotations.contains(Annotation.ALL_BROADCASTS)) {
```
dragospitica: I suggest putting this if statement after computing `all` (line 613) since previous computations are redundant if this is true

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -594,40 +602,53 @@ private void mergeFilmProperties(Application application, Film chosen,
...
             Publisher sourceForBroadcasts = Iterables.getOnlyElement(first).getSource();
-            chosen.setBroadcasts(Sets.newHashSet(
-                    Iterables.concat(
-                            Iterables.transform(
-                                    Iterables.filter(all, isPublisher(sourceForBroadcasts)),
-                                    Item::getBroadcasts
+            chosen.setBroadcasts(
+                    all.stream()
+                            .filter(item ->
+                                    activeAnnotations.contains(Annotation.ALL_MERGED_BROADCASTS)
```
dragospitica: if activeAnnotations contains the `ALL_MERGED_BROADCASTS`, it means no broadcasts will be filtered out. This seems similar to what the other annotation `ALL_BROADCASTS` does no?

EmiEnache: Yes it is similar, but there's the other step below it (the matchAndMerge for loop) that makes the difference.

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
EmiEnache: Good catch, done

PR #605, opened 2019-05-30
Support API reads at a higher consistency level 

dragospitica:
when using new request parameter and access role

JIRA ticket https://metabroadcast.atlassian.net/browse/ENG-320

### DIFF
--- a/atlas-core/src/main/java/org/atlasapi/equivalence/AnnotationBasedMergingEquivalentsResolver.java
+++ b/atlas-core/src/main/java/org/atlasapi/equivalence/AnnotationBasedMergingEquivalentsResolver.java
@@ -1,10 +1,12 @@
 package org.atlasapi.equivalence;
 
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.function.Predicate;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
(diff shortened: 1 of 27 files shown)


atlas-core/src/main/java/org/atlasapi/equivalence/EquivalentsResolver.java
```diff
@@ -21,12 +21,22 @@
      * @param ids               - requested numeric keys of equivalent content.
      * @param selectedSources   - sources of the equivalent set to resolve.
      * @param activeAnnotations - components of the model to resolve.
+     * @param isHigherReadConsistencyQuery - boolean request parameter which decides the read
```
EmiEnache: Description could be clearer -  it doesn't decide the read consistency level in the (implied) sense that it tells you what it is set to: it just sets to QUORUM is true (correct me if I am wrong right). Otherwise leaves it whatever is set/default. 

So I would say description of should be "boolean request parameter which sets the read consistency level to QUORUM" maybe.

atlas-core/src/main/java/org/atlasapi/equivalence/AnnotationBasedMergingEquivalentsResolver.java
```diff
@@ -62,6 +83,18 @@ public AnnotationBasedMergingEquivalentsResolver(EquivalentsResolver<E> resolver
         }
     }
 
+    private boolean isHigherReadConsistencyQuery(
+            AttributeQuerySet operands,
+            AccessRoles accessRoles
+    ) {
+        return !operands.isEmpty() && operands.stream().anyMatch(isHigherReadConsistency())
```
tristanbeardmore-gray: Don't think you need the isEmpty check here, but if you do make the logic clearer here with bracketing since you're combining && and ||

PR #614, opened 2019-11-20
Eng 548 - Adding is_unpublished annotation for content

obroomhall:
You can now use is_unpublished annotation when hitting the /4/content endpoint.
If called without non_merged annotation, will return error as unimplemented.
Otherwise, will return content as before, but with an is_published flag to signify whether or not the content is actively published.


atlas-cassandra/src/main/java/org/atlasapi/content/CassandraEquivalentContentStore.java
```diff
@@ -442,7 +444,25 @@ public void onFailure(Throwable t) {
                         Entry::getKey
                 ));
 
+        boolean allowUnpublished = activeAnnotations.contains(Annotation.IS_PUBLISHED);
```
EmiEnache: Is this annotation meant to return unpublished content? Because if so, should it not be called "IS_UNPUBLISHED"?

PR #616, opened 2020-01-06
Eng 559

EmiEnache:


atlas-cassandra/src/main/java/org/atlasapi/content/v2/serialization/setters/EpisodeSetter.java
```diff
@@ -52,9 +61,35 @@ public void deserialize(org.atlasapi.content.Content content, Content internal)
...
+                                            "Episode with id %s has more than one series ref",
+                                            episodeId
+                                    )
+                            )
+                            .setSource("atlas-deer")
+                            .setSeverity(Severity.WARNING)
+                            .setTimestamp(OffsetDateTime.now())
+                            .build();
+                    pagerDutyClient.trigger(payload, String.valueOf(episodeId));
+                } catch (NullPointerException e) {
```
tristanbeardmore-gray: Just catch all exceptions here

atlas-cassandra/src/main/java/org/atlasapi/content/v2/serialization/setters/EpisodeSetter.java
```diff
@@ -52,9 +61,35 @@ public void deserialize(org.atlasapi.content.Content content, Content internal)
...
+                    log.warn("Cannot throw PagerDuty alert because no config found for it.");
```
tristanbeardmore-gray: Can make this log warning more generic for any exception, and add `e` as the last parameter to log the stacktrace

atlas-cassandra/src/main/java/org/atlasapi/content/v2/serialization/setters/EpisodeSetter.java
```diff
@@ -52,9 +61,35 @@ public void deserialize(org.atlasapi.content.Content content, Content internal)
 
         Map<Ref, SeriesRef> seriesRefs = internal.getSeriesRefs();
         if (seriesRefs != null && !seriesRefs.isEmpty()) {
-            if(seriesRefs.size() > 1) {
-                log.error("Episode with id {} has more than one series ref", internal.getId());
+
+            if (seriesRefs.size() > 1) {
+                log.error("Episode with id {} has more than one series ref", episodeId);
+                try {
+                    pdIntegrationKey = System.getenv("PD_INTEGRATION_KEY"); // NPE if missing
```
tristanbeardmore-gray: That won't throw an exception itself. If you want to have it throw there use `checkNotNull`

atlas-cassandra/src/main/java/org/atlasapi/content/v2/serialization/setters/EpisodeSetter.java
```diff
@@ -22,6 +27,9 @@
     private final SeriesRefSerialization seriesRefSerialization = new SeriesRefSerialization();
     private final RefSerialization refSerialization = new RefSerialization();
 
+    private String pdIntegrationKey = null;
```
tristanbeardmore-gray: You don't need these

atlas-cassandra/src/main/java/org/atlasapi/content/v2/serialization/setters/EpisodeSetter.java
EmiEnache: System.getenv throws the NPE itself if the var it reads is null, so won't make this change.

PR #631, opened 2020-03-30
Add location provider to model + annotation to expose them

EmiEnache:


atlas-core/src/main/java/org/atlasapi/content/Provider.java
```diff
@@ -0,0 +1,64 @@
...
+public class Provider {
+
+    private String name;
+    private String iconUrl;
+
+    public Provider() {
+    }
+
+    public Provider(@NotNull String name, @Nullable String iconUrl) {
+        this.name = checkNotNull(name);
```
tristanbeardmore-gray, line 19: You've added a checkNotNull to this particular constructor whilst allowing it in theory to be null in most others - it would probably be good to be consistent in either one or the other.

Might also be worth checking if checkNotNulls are ever used in model classes, or if null checking is done outside of these classes, such as in the serialization/deserialization steps.

PR #634, opened 2020-04-09
Fix lowest id messing source precedence data

EmiEnache:
(branch name is misleading)

This is to solve a merging logic bug where data from highest precedence source gets overwritten because of a weird, hacky coding related to showing lowest ID (and its publisher) in the merged result. See ENG-675, but basically this will fix the bug related to images only - rest of fields are out of scope because we'd rather rework this bit of the pipeline instead of fixing each field one by one.


EmiEnache:
These changes would introduce another bug which can't easily be solved without reworking the way merging is done. It was decided that this work should be abandoned in favour of that (see Jira / ENG-820).

PR #635, opened 2020-04-22
Sherlock search

obroomhall:
Adds sherlock searching endpoint


atlas-api/Dockerfile
```diff
@@ -41,6 +41,10 @@ CMD java \
     -Delasticsearch.index="$ELASTICSEARCH_INDEX" \
     -Delasticsearch.requestTimeout="$ELASTICSEARCH_REQUESTTIMEOUT" \
     -Delasticsearch.seeds="$ELASTICSEARCH_SEEDS" \
+    -Dsherlock.scheme="$SHERLOCK_SCHEME" \
+    -Dsherlock.hostname="$SHERLOCK_HOSTNAME" \
+    -Dsherlock.port="$SHERLOCK_PORT" \
+    -Dsherlock.timeout.milliseconds="$SHERLOCK_TIMOUT_MILLIS" \
```
tristanbeardmore-gray, line 47: TIMEOUT?

atlas-api/src/main/java/org/atlasapi/query/QueryWebModule.java
```diff
@@ -935,7 +957,76 @@ SearchController searchController() {
...
                         requestWriter(),
                         channelGroupResolver,
-                        idCodec()
+                        idCodec
+                )
+        );
+    }
+
+    private List<SherlockAttribute<?, ?, ?>> sherlockAttributes() {
+        ContentMapping CONTENT_MAPPING = IndexMapping.getContentMapping();
```
tristanbeardmore-gray, line 966: Normally we only use all caps for static values.

atlas-api/src/main/java/org/atlasapi/query/QueryWebModule.java
```diff
@@ -935,7 +957,76 @@ SearchController searchController() {
...
+                ))
+                .add(new SherlockAttribute<Boolean, Instant, DateMapping>(
+                        SherlockAttributes.SCHEDULE_UPCOMING_PARAM,
+                        CONTENT_MAPPING.getBroadcasts().getTransmissionStartTime(),
+                        BooleanCoercer.create()
+                ) {
+                    @Override protected NamedParameter<Instant> createParameter(
+                            DateMapping mapping, Boolean value
+                    ) {
+                        if (value) {
```
tristanbeardmore-gray, line 991: If value can be null this will probably throw an NPE

atlas-api/src/main/java/org/atlasapi/query/v5/search/ContentQueryResultWriter.java
```diff
@@ -0,0 +1,150 @@
...
+    private OutputContext outputContext(QueryContext queryContext) {
+
+        String channelGroupParam = queryContext.getRequest()
+                .getParameter(SherlockAttributes.SCHEDULE_CHANNEL_GROUP_PARAM);
+
+        OutputContext.Builder builder = OutputContext.builder(queryContext);
+
+        List<String> channelGroups;
+        if (!Strings.isNullOrEmpty(channelGroupParam)) {
+            channelGroups = Arrays.asList(channelGroupParam.split("\\s*,\\s*"));
```
tristanbeardmore-gray, line 89: Is this kind of thing done elsewhere? Seems a bit unusual but won't debate it if that's how we treat other list parameters.

atlas-api/src/main/java/org/atlasapi/query/v5/search/ContentQueryResultWriter.java
```diff
@@ -0,0 +1,150 @@
...
+
+        List<ChannelGroup<?>> channelGroups = new ArrayList<>();
+
+        for (String idParam : idParams) {
+            Id id = Id.valueOf(codec.decode(idParam));
+
+            Optional<ChannelGroup<?>> channelGroupOptional;
+            try {
+                channelGroupOptional = Futures.getChecked(
+                        channelGroupResolver.resolveIds(ImmutableList.of(id)),
```
tristanbeardmore-gray, line 132: Why are we querying these one by one when the method allows us to pass in multiple? This will be a significant bottleneck if someone wants to specify a large number of channel groups.

atlas-api/src/main/java/org/atlasapi/query/v5/search/ContentResolvingSearcher.java
```diff
@@ -0,0 +1,103 @@
...
+            QueryContext queryContext
+    ) {
+        List<Id> ids = decodeIds(encodedIds);
+        if (ids.isEmpty()) {
+            return Futures.immediateFuture(ResolvedEquivalents.empty());
+        } else {
+            return contentResolver.resolveIds(
+                    ids,
+                    queryContext.getApplication(),
+                    queryContext.getAnnotations().all(),
```
tristanbeardmore-gray, line 78: I suspect this works but I think there's a bit of risk in doing this. For example we recently started throwing errors with the is_published annotation if non_merged was not specified as well - if we do other kinds of validation, such as not allowing two annotations to be used at the same time it will cause issues here.

I don't necessarily have a good solution for this, it's more to highlight the risk - it might be good within the Annotation enum to have a static list of all annotations for use by the searcher, which we keep up to date as we add new fields to the api.

atlas-api/src/main/java/org/atlasapi/query/v5/search/SearchController.java
```diff
@@ -0,0 +1,189 @@
...
+                queryBuilder = SearchQuery.builder();
+            } else {
+                queryBuilder = SearchQuery.getDefaultQuerySearcher(query);
+            }
+
+            for (NamedParameter<?> parameter : parameters) {
+                if (parameter instanceof SearchParameter) {
+                    queryBuilder.addSearcher((SearchParameter)parameter);
+                } else {
+                    queryBuilder.addFilter((FilterParameter<?>)parameter);
```
tristanbeardmore-gray, line 123: Is this a safe cast? It may be worth doing a check as an else if so that if you add more in future it won't cause this to break.

atlas-api/src/main/java/org/atlasapi/query/v5/search/SearchController.java
```diff
@@ -0,0 +1,189 @@
...
+            String name = attribute.getParameterName();
+            if (parameterMap.containsKey(name)) {
+
+                List<String> values = Arrays.stream(parameterMap.get(name))
+                        .map(SPLITTER::splitToList)
+                        .flatMap(Collection::stream)
+                        .distinct()
+                        .collect(Collectors.toList());
+
+                if (values.contains(EXISTS_KEYWORD)) {
```
tristanbeardmore-gray, line 167: This implementation of exists and not exists feels a bit off to me, especially because this technically allows you to pass in a list of legitimate values, the exists keyword, and the non exists keyword. Perhaps some validation would be worthwhile, or even a different parameter that is uses to specify which fields should and should not exist?

atlas-api/src/main/java/org/atlasapi/query/v5/search/coercer/InstantRangeCoercer.java
```diff
@@ -0,0 +1,34 @@
...
+        super(Instant::parse);
+    }
+
+    public static InstantRangeCoercer create() {
+        return new InstantRangeCoercer();
+    }
+
+    @Override
+    protected Instant[] bisectAndCoerce(String value) throws InvalidAttributeValueException {
+        if (value.length() % 2 != 1) { // Instant range values must be in the same format
```
tristanbeardmore-gray, line 20: This doesn't really tell you they're the same format, just that the combined length is even.

tristanbeardmore-gray: Also will you need to trim spaces at all for any of these coercers?

atlas-api/src/main/java/org/atlasapi/query/v5/search/coercer/InstantRangeCoercer.java
```diff
@@ -0,0 +1,34 @@
...
+            throw new InvalidAttributeValueException(value);
+        }
+
+        int mid = (value.length()-1)/2;
```
tristanbeardmore-gray, line 24: Is there any reason for not using a separator character for this coercer unlike the others?

atlas-api/src/main/java/org/atlasapi/query/v5/search/coercer/RangeCoercer.java
```diff
@@ -0,0 +1,52 @@
...
+            throw new InvalidAttributeValueException(value);
+        } else if (value.startsWith(RANGE_SEPARATOR)) {
+            return new Range<>(null, coercerFunction.apply(value.substring(1)));
+        } else if (value.endsWith(RANGE_SEPARATOR)) {
+            return new Range<>(coercerFunction.apply(value.substring(0, value.length()-1)), null);
+        } else {
+            try {
+                T parsedValue = coercerFunction.apply(value);
+                return new Range<>(parsedValue, parsedValue);
+            } catch (Exception e) {
```
tristanbeardmore-gray, line 32: I wouldn't use exceptions for this, it's kind of abusing what they're for. You could have a method return a particular value (null or empty optional) etc to indicate that you weren't able to extract a single value instead.

atlas-api/src/main/java/org/atlasapi/query/v5/search/coercer/InstantRangeCoercer.java
obroomhall: Spaces are trimmed in the SearchController, updated the InstantRangeCoercer to only accept yyyy-MM-ddTHH:mm:ssZ-yyyy-MM-ddTHH:mm:ssZ

PR #641, opened 2020-05-21
Rework merging pipeline: eng-820

EmiEnache:
See ticket for full context, but basically merging doesn't work well for every field if lowest ID is not from highest precedence source. This is due to some hacky code which tampered with an existing piece of context from the equivset to create the merged content. 

This rework instead aims to create a new piece of content from scratch to represent the merged content, while maintaining all existing requirements and constraints.

https://metabroadcast.atlassian.net/browse/ENG-820

### DIFF
--- a/atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
+++ b/atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
@@ -119,49 +119,6 @@ public class OutputContentMerger implements EquivalentsMergeStrategy<Content> {
         this.hierarchyChooser = checkNotNull(hierarchyChooser);
     }
 
-    @Deprecated
-    public <T extends Described> List<T> merge(
-            Application application,
-            List<T> contents,
-            Set<Annotation> activeAnnotations
-    ) {
-
-        Ordering<Sourced> publisherComparator = application.getConfiguration()
-                .getReadPrecedenceOrdering()
-                .onResultOf(Sourceds.toPublisher());
-
-        List<T> merged = Lists.newArrayListWithCapacity(contents.size());
-        Set<T> processed = Sets.newHashSet();
-
-        for (T content : contents) {
-            if (processed.contains(content)) {
-                continue;
-            }
-            List<T> orderedContent = publisherComparator.sortedCopy(findSame(content, contents));
-            processed.addAll(orderedContent);
-
-            T chosen = createMerged(orderedContent);
-
-            // defend against broken transitive equivalence
-            if (merged.contains(chosen)) {
-                continue;
-            }
-
-            //TODO remove chosen + notChosen concats.
-            if (chosen instanceof Container) {
-                mergeIn(application, (Container) chosen, filterEquivs(orderedContent, Container.class));
-            }
-            if (chosen instanceof Item) {
-                mergeIn(application, (Item) chosen, filterEquivs(orderedContent, Item.class), activeAnnotations);
-            }
-            if (chosen instanceof ContentGroup) {
-                mergeIn(application, (ContentGroup) chosen, filterEquivs(orderedContent, ContentGroup.class));
-            }
-            merged.add(chosen);
-        }
-        return merged;
-    }
-
     //sortedEquivalents must be ordered according to source precedence
     private <T extends Described> T createMerged(Iterable<? extends T> sortedEquivalents) {
 
@@ -233,10 +190,12 @@ public class OutputContentMerger implements EquivalentsMergeStrategy<Content> {
     }*/
 
     @Override
-    public <T extends Content> T merge(T chosen, final Iterable<? extends T> sortedEquivalents,
+    public <T extends Content> T merge(
+            final Iterable<? extends T> sortedEquivalents,
             final Application application,
-            Set<Annotation> activeAnnotations) {
-        chosen = createMerged(sortedEquivalents);
+            Set<Annotation> activeAnnotations)
+    {
+        T chosen = createMerged(sortedEquivalents);
         return chosen.accept(new ContentVisitorAdapter<T>() {
 
             @Override
@@ -447,7 +406,7 @@ public class OutputContentMerger implements EquivalentsMergeStrategy<Content> {
                 Content::getCountriesOfOrigin
         ));
 
-        //TODO add mergePeople()
+        //TODO add mergePeople() (have only people from highest precedence content)
     }
 
     private <T extends Content> void mergeDuration(T chosen, Iterable<T> orderedContent) {
@@ -572,6 +531,7 @@ public class OutputContentMerger implements EquivalentsMergeStrategy<Content> {
         }
     }
 
+    //TODO have all images ordered by source precedence, not just higher
     private <T extends Described> void applyImagePrefs(Application application, T chosen,
             Iterable<T> orderedContent) {
         if (application.getConfiguration().isImagePrecedenceEnabled()) {
--- a/atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
+++ b/atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
@@ -3,7 +3,6 @@ package org.atlasapi.output;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.HashSet;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -35,6 +34,7 @@ import org.atlasapi.content.LocalizedTitle;
 import org.atlasapi.content.LocationSummary;
 import org.atlasapi.content.ReleaseDate;
 import org.atlasapi.content.Series;
+import org.atlasapi.content.Song;
 import org.atlasapi.content.Subtitles;
 import org.atlasapi.content.Tag;
 import org.atlasapi.entity.Award;
@@ -140,7 +140,7 @@ public class OutputContentMerger implements EquivalentsMergeStrategy<Content> {
             List<T> orderedContent = publisherComparator.sortedCopy(findSame(content, contents));
             processed.addAll(orderedContent);
 
-            T chosen = createChosen(orderedContent);
+            T chosen = createMerged(orderedContent);
 
             // defend against broken transitive equivalence
             if (merged.contains(chosen)) {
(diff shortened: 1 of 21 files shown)


gorgeoux:
Hi Toby,

Is this something that you need to see or did you get a copy anyway.  Just
thought I would check.

Best wishes,

Meriel

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -117,124 +120,79 @@ public OutputContentMerger(EquivalentSetContentHierarchyChooser hierarchyChooser
...
+        IdAndPublisher idAndPublisher = findLowestIdContent(sortedEquivalents);
+        Id id = idAndPublisher.getId();
+        merged.setId(id);
+        Publisher publisher = idAndPublisher.getPublisher();    // update publisher accordingly
+        merged.setPublisher(publisher);
+
+        // Remove self from same_as set
+        Set<EquivalenceRef> equivs = Sets.newHashSet(merged.getEquivalentTo());
+        equivs.remove(EquivalenceRef.valueOf(merged));
+        merged.setEquivalentTo(equivs);
```
tristanbeardmore-gray, line 153: The highest precedent content will be missing from the equiv set if it is different to idAndPublisher I believe.

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -645,38 +579,32 @@ private void mergeFilmProperties(Application application, Film chosen,
...
             );
             return;
         }
 
-        List<T> notChosenOrdered = application.getConfiguration()
-                .getReadPrecedenceOrdering()
-                .onResultOf(Sourceds.toPublisher())
-                .sortedCopy(notChosen);
-
+        //TODO remove this ordering, as orderedContent is already sorted by source precedence at this point
```
tristanbeardmore-gray, line 601: Is this something you still need to do?

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -726,7 +654,7 @@ private void mergeFilmProperties(Application application, Film chosen,
 
         if (chosen.getBroadcasts() != null && !chosen.getBroadcasts().isEmpty()) {
             for (Broadcast chosenBroadcast : chosen.getBroadcasts()) {
-                matchAndMerge(chosenBroadcast, notChosenOrdered);
+                matchAndMerge(chosenBroadcast, Lists.newArrayList(orderedContent));
```
tristanbeardmore-gray, line 657: If you need to use a list here for the second parameter create it outside the for loop to prevent recreating it each time.

tristanbeardmore-gray: matchAndMerge could be changed to accept an iterable instead of a list by the looks of the code which will prevent needing to create a list anyway.

atlas-core/src/main/java/org/atlasapi/output/StrategyBackedEquivalentsMerger.java
```diff
@@ -26,6 +25,8 @@
         implements ApplicationEquivalentsMerger<E> {
 
     private final EquivalentsMergeStrategy<E> strategy;
+    @VisibleForTesting public static final Ordering<Equivalable> ID_ORDERING =
```
alianos-, line 28: Do you access this via a test somewhere?

atlas-core/src/main/java/org/atlasapi/output/StrategyBackedEquivalentsMerger.java
```diff
@@ -63,18 +64,15 @@ public StrategyBackedEquivalentsMerger(EquivalentsMergeStrategy<E> strategy) {
...
+                sortedEquivalents.add(0, requested.get());
             }
         }
 
-        Iterable<T> notChosen = Iterables.filter(
-                sortedEquivalents,
-                Predicates.not(idIs(chosen.getId()))
-        );
-        return ImmutableList.of(strategy.merge(chosen, notChosen, application, activeAnnotations));
+        return ImmutableList.of(strategy.merge(sortedEquivalents, application, activeAnnotations));
```
alianos-, line 72: This seems exactly the same as with the trivialMerge now, so either trivialMerge can be removed, or trivialMerged was some optimisation that we can adapt.

atlas-core/src/main/java/org/atlasapi/output/StrategyBackedEquivalentsMerger.java
```diff
@@ -63,18 +64,15 @@ public StrategyBackedEquivalentsMerger(EquivalentsMergeStrategy<E> strategy) {
 
             if (requested.isPresent()
```
alianos-, line 65: please leave a comment about what this does, also mentioning that at the time of the rework it is unclear if anything relies on it, but that nothing built in the last 3 years relies on it. And that it is the view of the current team that this is wrong and should be removed as it causes the result of the merge to be different for different IDs of the equiv set.

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
EmiEnache: It shouldn't be in the equivalentTo set if that's the case though, should it? What's the issue here

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
EmiEnache: I want to push this out sooner rather than later, and don't want to spend time ensuring that removing the ordering is indeed safe - so just left this note for a later time when we work on efficiency.

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -117,124 +120,79 @@ public OutputContentMerger(EquivalentSetContentHierarchyChooser hierarchyChooser
...
-            // defend against broken transitive equivalence
-            if (merged.contains(chosen)) {
-                continue;
-            }
+        // Finally, we need the lowest id in the equiv set, in an attempt to keep the id stable
+        IdAndPublisher idAndPublisher = findLowestIdContent(sortedEquivalents);
+        Id id = idAndPublisher.getId();
+        merged.setId(id);
+        Publisher publisher = idAndPublisher.getPublisher();    // update publisher accordingly
+        merged.setPublisher(publisher);
```
alianos-, line 148: Styling detail, consider doing the above without creating new objects for id and publisher (the compiler will remove them anyway but it will read better).

atlas-core/src/main/java/org/atlasapi/output/StrategyBackedEquivalentsMerger.java
EmiEnache: Yes

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
alianos-: This doesn't sound right in general. Why don't we copy the equiv set of the lowest ID instead of trying to recreate it?

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -117,124 +120,79 @@ public OutputContentMerger(EquivalentSetContentHierarchyChooser hierarchyChooser
...
+        T merged = (T) mostSpecificTypeContent.createNew();
 
-            chosen = mergeIdAndEquivTo(chosen);
+        // Then we need as much data as possible from the highest precedence source possible
+        merged = highestPrecedenceContent.copyToPreferNonNull(merged);
 
-            // defend against broken transitive equivalence
-            if (merged.contains(chosen)) {
-                continue;
-            }
```
alianos-, line 146: do we know what this means, and have defended against it?

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -117,124 +120,79 @@ public OutputContentMerger(EquivalentSetContentHierarchyChooser hierarchyChooser
...
 
-            Map.Entry<T, List<T>> chosenAndRest = getChosenAndRest(same);
-            T chosen = chosenAndRest.getKey();
-            List<T> notChosen = chosenAndRest.getValue();
+        //Unchecked casting is safe here because we get the most specific type from the hierarchy
+        T merged = (T) mostSpecificTypeContent.createNew();
 
-            chosen = mergeIdAndEquivTo(chosen);
+        // Then we need as much data as possible from the highest precedence source possible
+        merged = highestPrecedenceContent.copyToPreferNonNull(merged);
```
alianos-, line 138: would it make sense to not do this, and let the merging logic fill it in as with any other item of the list? That would be a lot cleaner in my brain.

tristanbeardmore-gray: I think a caveat might be that if a field has no merging logic rules then it won't be filled in at all, instead of at least being filled in when being initially copied.

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
EmiEnache: I get what you meant now Tristan, and you're right.
As per Andreas' point I'll try to use lowest ID equiv set instead.

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
EmiEnache: What Tristan is saying is the reason this is here. But it does make things less efficient, so it does have a downside. Adding it would still miss cases where data is not in highestPrecedence but instead in some lower precedence source content. 

How should we leave it @alianos- ?

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
EmiEnache: I believe this was related to the other method which had 'merged' potentially by a List of content, which was not used anywhere except tests, so I thought it safe to remove this.

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
EmiEnache: https://github.com/atlasapi/atlas-deer/commit/cb5c5cbfa40971881d1b7271b35071eb4e7c17f0

atlas-core/src/main/java/org/atlasapi/output/StrategyBackedEquivalentsMerger.java
EmiEnache: I'm hesitant to remove as trivialMerge wasn't much different before than it is now tbh, and I don't know why it's there unless it's just a very specific optimisation - and if so then what should we changed about it?

atlas-core/src/main/java/org/atlasapi/output/StrategyBackedEquivalentsMerger.java
EmiEnache: Done, see https://github.com/atlasapi/atlas-deer/commit/f3c695e04cbc3529e04b002971d31756688d7da0

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
alianos-: yeap thats fine. Can't say I prefer it but it is what it is and at least it has some function.

PR #645, opened 2020-06-02
Fix merging logic for languages & certificates

EmiEnache:
This adds merging logic for languages & certificates
to all content types (previously, it was just films). 
This fixes cases such as editors renaming a brand, 
which adds a higher precedence source content
with no language the equiv set, resulting in no
language being visible in Shaman. See ENG-813.

### DIFF
--- a/atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
+++ b/atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
@@ -612,7 +612,6 @@ public class OutputContentMerger implements EquivalentsMergeStrategy<Content> {
         chosen.setCertificates(combinedCertificates);
     }
 
-    // find first non-null, non-empty set of languages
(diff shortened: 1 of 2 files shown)


atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -606,6 +603,25 @@ private void mergeFilmProperties(Application application, Film chosen,
...
+    private <T extends Content> void mergeCertificates(T chosen, Iterable<T> orderedContent) {
+        Set<Certificate> combinedCertificates = MoreStreams.stream(orderedContent)
+                .map(Content::getCertificates)
+                .flatMap(Collection::stream)
+                .collect(Collectors.toSet());
+
+        chosen.setCertificates(combinedCertificates);
+    }
+
+    // find first non-null, non-empty set of languages
```
alianos-, line 615: the comment is out of date

PR #649, opened 2020-07-02
Add retries to select queries in CassandraEquivalentContentStore...

tristanbeardmore-gray:
...if the set of setIds from the graph and data queries do not match
to try and work around an unavoidable race condition which otherwise
led to an error.

### DIFF
--- a/atlas-cassandra/src/main/java/org/atlasapi/content/CassandraEquivalentContentStore.java
+++ b/atlas-cassandra/src/main/java/org/atlasapi/content/CassandraEquivalentContentStore.java
@@ -1,46 +1,5 @@
 package org.atlasapi.content;
 
-import java.io.IOException;
-import java.nio.ByteBuffer;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Map.Entry;
-import java.util.Set;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Executor;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


atlas-cassandra/src/main/java/org/atlasapi/content/CassandraEquivalentContentStore.java
```diff
@@ -767,55 +764,93 @@ private boolean containedInGraph(Id contentId, java.util.Optional<EquivalenceGra
...
+                List<Row> graphResults = getGraphResults(combinedResults);
+                List<Row> dataResults = getDataResults(combinedResults);
+                boolean resultsMatch = resultsMatch(graphResults, dataResults);
+                if (resultsMatch || attempt.incrementAndGet() > NUMBER_OF_CASSANDRA_SELECT_RETRIES) {
+                    if (!resultsMatch) {
+                        log.warn("Exceeded retry count for combined data and graph future");
+                    }
+                    return Futures.immediateFuture(new GraphAndDataResults(graphResults, dataResults));
+                } else {
+                    return combineFuture(dataStatement, graphStatement);
```
alianos-, line 809: do you want to log a line here so that we get an idea how often it is happening?

alianos-: also should we implement an incremental backoff? Maybe put the logging in, test, then try with a backoff see if helped.

tristanbeardmore-gray: I thought about it but wasn't too sure if it would help since if there were a mismatch then presumably the table has already been updated. Not against the idea though just in case.

Happy to add an extra log line - I wanted to include the content id in the other log line but it would require passing that information around a bit so decided against it for the time being as well.

PR #661, opened 2020-09-07
Use isCollectionOfValues field to determine value splitting logic

obroomhall:
Workaround specifically for the q parameter. Could use isCollectionOfValues on the Attribute object for a more general solution, however, the field has not been in use and thus every parameter is currently treated as a list, so as not to cause any unwanted side effects of correctly implementing the field, do this instead.

### DIFF
--- a/atlas-api/src/main/java/org/atlasapi/query/common/attributes/QueryAttributeParser.java
+++ b/atlas-api/src/main/java/org/atlasapi/query/common/attributes/QueryAttributeParser.java
@@ -1,6 +1,7 @@
 package org.atlasapi.query.common.attributes;
 
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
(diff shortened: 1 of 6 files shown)


atlas-core/src/main/java/org/atlasapi/criteria/attribute/Attributes.java
```diff
@@ -84,12 +84,12 @@ private Attributes() {
             "topics.relationship",
             Identified.class
     );
-    public static final Attribute<Float> TAG_WEIGHTING = FloatAttribute.single(
```
alianos-, line 87: That doesn't make sense. This can now accept multiple values where it should only accept one, defeating the whole point of the validation.

atlas-api/src/main/java/org/atlasapi/query/common/attributes/QueryAttributeParser.java
```diff
@@ -68,13 +69,22 @@ public AttributeQuerySet parse(HttpServletRequest request) throws QueryParseExce
...
+                QueryAtomParser<?> parser = parsers.get(attribute);
 
                 for (String paramString : param.getValue()) {
-                    operands.add(parser.parse(param.getKey(), splitVals(paramString)));
+
+                    Iterable<String> values;
+                    if (attribute.isCollectionOfValues()) {
+                        values = splitVals(paramString);
+                    } else {
+                        values = Collections.singletonList(paramString);
```
alianos-, line 84: this will now effectively not verify that you have been given a single parameter, but treat the parameter as single.

atlas-core/src/main/java/org/atlasapi/criteria/attribute/Attributes.java
obroomhall: The field that `.single()` and `.list()` set is unused, so all parameters are already treated as being lists of values.

obroomhall: I have reverted these `.single()` to `.list()` changes to keep the code as is, and added a `.singleFreeText()` method for use by the Q parameter.

atlas-api/src/main/java/org/atlasapi/query/common/attributes/QueryAttributeParser.java
obroomhall: Given recent changes, I don't think this comment is relevant anymore, but please let me know if it is still an issue

PR #662, opened 2020-10-09
Fix cassandra consistency level role regex pattern

EmiEnache:


EmiEnache:
Originally written by Tristan but had to be set aside, so today I tested it (after discussing it with Tristan)

PR #681, opened 2021-05-20
Prevent c4 film-as-episodes messing with hierarchy in merged result

EmiEnache:


atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -68,6 +71,7 @@
 
     private static final long BROADCAST_START_TIME_TOLERANCE_IN_MS = Duration.standardMinutes(5)
             .getMillis();
+    private static final String DO_NOT_FILL_IN_FILM_HIERARCHY_FROM_EPISODES = "do-not-fill-in-film-hierarchy-from-episodes";
```
tristanbeardmore-gray, line 74:
```suggestion
    private static final String DO_NOT_MERGE_FILM_HIERARCHY_FROM_EPISODES = "do-not-merge-film-hierarchy-from-episodes";
```

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -406,12 +411,24 @@ private T uncheckedCast(Content c) {
...
+            Application application,
+            T chosen,
+            Iterable<T> orderedContent,
+            boolean chosenIsFilm
+    ) {
         if (chosen.getContainerRef() != null) {
             return;
         }
         boolean chosenIsEpisode = chosen instanceof Episode;
+        boolean ignoreEpisodeHierarchies = chosenIsFilm
```
alianos-, line 424: I think this variable could be calculated outside the method and thus reduce the additional argument to 1 instead of 2.

PR #699, opened 2021-11-15
ENG-1128

IvanKaraslavov:
Adds new query filters for BT called dtt_channels_only and ip_channels only which filter channel groups based on their transmission type. The filters are added both to the channel groups and content endpoints.


atlas-core/src/main/java/org/atlasapi/criteria/attribute/Attributes.java
```diff
@@ -112,14 +112,22 @@ private Attributes() {
...
-    public static final Attribute<Id> CHANNEL_GROUP_DTT_CHANNELS = IdAttribute.list(
+    public static final Attribute<Id> CHANNEL_GROUP_IP_CHANNELS_V2 = IdAttribute.list(
+            "ip_channels_only",
+            ChannelGroup.class
+    );
+    public static final Attribute<Id> CHANNEL_GROUP_DTT_CHANNELS_V1 = IdAttribute.list(
             "dtt_only",
             ChannelGroup.class
     );
+    public static final Attribute<Id> CHANNEL_GROUP_DTT_CHANNELS_ONLY_V2 = IdAttribute.list(
```
tristanbeardmore-gray, line 127:
```suggestion
    public static final Attribute<Id> CHANNEL_GROUP_DTT_CHANNELS_V2 = IdAttribute.list(
```

PR #703, opened 2021-12-07
ENG-1149

IvanKaraslavov:
Adds release dates on content level so all types of content can have them.


atlas-api/src/main/java/org/atlasapi/generation/generated/model/IdentifiedInfo.java
```diff
@@ -181,6 +181,16 @@
                             .withJsonType(JsonType.MAP)
                             .build()
             )
+            .add(
+                    FieldInfo.builder()
```
tristanbeardmore-gray, line 185: I believe technically this shouldn't be added in this class but to ContentInfo because this field doesn't exist at the Identified level. But it should also be added to the other *Info classes that correspond to the subclasses of Content because I remember that being done for most fields.

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -732,6 +730,15 @@ private void mergeFilmProperties(Application application, Film chosen,
         chosen.setCertificates(combinedCertificates);
     }
 
+    private <T extends Content> void mergeReleaseDates(T chosen, Iterable<T> orderedContent) {
+        Set<ReleaseDate> combinedReleaseDates = MoreStreams.stream(orderedContent)
+                .map(Content::getReleaseDates)
+                .flatMap(Collection::stream)
+                .collect(Collectors.toSet());
```
tristanbeardmore-gray, line 737:
```suggestion
                .collect(MoreCollectors.toImmutableSet());
```
To preserve order and match existing logic.

PR #704, opened 2021-12-14
Kafka upgrade batch 1 part 2

pawel-mtbcst:


atlas-core/src/main/java/org/atlasapi/messaging/JacksonMessageSerializer.java
```diff
@@ -168,6 +169,7 @@ private static ObjectMapper mapper() {
         mapper.registerModule(new GuavaModule());
         mapper.registerModule(new AtlasModelModule());
         mapper.registerModule(new MessagingModule());
+        mapper.registerModule(new JavaTimeModule());
```
tristanbeardmore-gray, line 172: Is there a need to add this? I thought your libraries don't use these serializers.

pawel-mtbcst: My bad, I was investigating whether we can use the migrated message on both Kafka brokers, but finally I decided it's simpler and safer t migrate both queues and once than doing this. Thanks

PR #705, opened 2021-12-16
ENG-1168

IvanKaraslavov:
Updates merging logic for release dates and restrictions

Release dates
1. Dedupe them by country + type fields
2. Add them in precedence order

Restrictions
1. Filter out all restrictions without meaningful data (their fields are null)
2. If the authorities of the remaining ones is null, set them as their content sources (for better visibility and merging)
3. If there are more than one restrictions from the same source with the same authority, pick the one with the highest minimal age
4. Dedupe them by the authority field
5. Add them in precedence order


atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -452,6 +463,24 @@ private T uncheckedCast(Content c) {
         }
     }
 
+    private <T extends Item> void mergeRestrictions(T chosen, Iterable<T> orderedContent) {
+        Map<String, Restriction> restrictionMap = new LinkedHashMap<>();
+        for (T content : orderedContent) {
+            Set<Restriction> restrictions = MoreStreams.stream(content.getRestrictions())
+                    .filter(this::containsMeaningfulData)
+                    // we are deduping on authority so if the restriction contains some information but the authority is not set, we are setting it as the content's source for better merging
+                    .peek(restriction -> {if(Strings.isNullOrEmpty(restriction.getAuthority())) {restriction.setAuthority(content.getSource().key());} })
```
pawel-mtbcst, line 472: It's hardly readable, extract it to seperate method and use like this:
`.peek(restriction  -> this.setAuthorityIfNecessary(restriction, content))`
or even:
`.peek(this::setAuthorityIfNecessary(content))`

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -981,3 +1010,27 @@ public Tag apply(Tag input) {
         }
     }
 }
+
+class ReleaseDateKey {
```
pawel-mtbcst, line 1014: Why not declare it as a private static inner class?
TBH I have never met whit this kind of class declaration other than as recruitment questions.

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -731,12 +760,12 @@ private void mergeFilmProperties(Application application, Film chosen,
...
-                .flatMap(Collection::stream)
-                .collect(MoreCollectors.toImmutableSet());
-
-        chosen.setReleaseDates(combinedReleaseDates);
+        Map<ReleaseDateKey, ReleaseDate> releaseDateMap = new LinkedHashMap<>();
+        for (T content : orderedContent) {
+            Set<ReleaseDate> releaseDates = content.getReleaseDates();
+            releaseDates.forEach(releaseDate -> releaseDateMap.putIfAbsent(new ReleaseDateKey(releaseDate), releaseDate));
+        }
+        chosen.setReleaseDates(releaseDateMap.values());
```
pawel-mtbcst, line 768: Here's the streaming version. I'm more used to streams, but I understand that it may not be as straightforward as your version, so I'm not insisting on it :) 

```
        chosen.setReleaseDates(StreamSupport.stream(orderedContent.spliterator(), false)
                .map(Content::getReleaseDates)
                .flatMap(Collection::stream)
                .collect(toMap(ReleaseDateKey::new, identity(), (r1, r2) -> r1))
                .values());
```

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -452,6 +463,35 @@ private T uncheckedCast(Content c) {
...
+    private <T extends Item> void mergeRestrictions(T chosen, Iterable<T> orderedContent) {
+        Map<String, Restriction> restrictionMap = new LinkedHashMap<>();
+        for (T content : orderedContent) {
+            MoreStreams.stream(content.getRestrictions())
+                    .filter(this::containsMeaningfulData)
+                    .peek(restriction -> this.setAuthorityIfNecessary(restriction, content))
+                    .sorted(Comparator.comparing(Restriction::getMinimumAge, Comparator.nullsFirst(Comparator.naturalOrder())).reversed())
+                    .forEach(restriction -> restrictionMap.putIfAbsent(restriction.getAuthority(), restriction));
+        }
+        chosen.setRestrictions(new LinkedHashSet<>(restrictionMap.values()));
```
tristanbeardmore-gray, line 475:
```suggestion
        chosen.setRestrictions(ImmutableSet.copyOf(restrictionMap.values()));
```

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -452,6 +463,35 @@ private T uncheckedCast(Content c) {
...
+    }
+
+    private <T extends Item> void setAuthorityIfNecessary(Restriction restriction, T content) {
+        // we are deduping on authority so if the restriction contains some information but the authority is not set, we are setting it as the content's source for better merging
+        if(Strings.isNullOrEmpty(restriction.getAuthority()))
+        {
```
tristanbeardmore-gray, line 481: Curly brackets are generally opened on the same line as their corresponding statement.

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -731,12 +771,12 @@ private void mergeFilmProperties(Application application, Film chosen,
...
-                .flatMap(Collection::stream)
-                .collect(MoreCollectors.toImmutableSet());
-
-        chosen.setReleaseDates(combinedReleaseDates);
+        Map<ReleaseDateKey, ReleaseDate> releaseDateMap = new LinkedHashMap<>();
+        for (T content : orderedContent) {
+            Set<ReleaseDate> releaseDates = content.getReleaseDates();
+            releaseDates.forEach(releaseDate -> releaseDateMap.putIfAbsent(new ReleaseDateKey(releaseDate), releaseDate));
+        }
+        chosen.setReleaseDates(releaseDateMap.values());
```
tristanbeardmore-gray, line 779:
```suggestion
        chosen.setReleaseDates(ImmutableMap.copyOf(releaseDateMap.values()));
```

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -452,6 +463,24 @@ private T uncheckedCast(Content c) {
...
+                    // we are deduping on authority so if the restriction contains some information but the authority is not set, we are setting it as the content's source for better merging
+                    .peek(restriction -> {if(Strings.isNullOrEmpty(restriction.getAuthority())) {restriction.setAuthority(content.getSource().key());} })
+                    .sorted(Comparator.comparing(Restriction::getMinimumAge).reversed())
+                    .collect(MoreCollectors.toImmutableSet());
+            restrictions.forEach(restriction -> restrictionMap.putIfAbsent(restriction.getAuthority(), restriction));
+        }
+        chosen.setRestrictions(new LinkedHashSet<>(restrictionMap.values()));
+    }
+
+    private boolean containsMeaningfulData(Restriction restriction) {
```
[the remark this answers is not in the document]
pawel-mtbcst: If you comment with Javadoc style, IDE will display it wherever you need it (ctrl+q).
Also, you can use annotations like @see that will be automatically converted to clickable links. 

```
/**
 * All content that should have no restriction instead have a meaningless restriction object.
 * In atlas-persistence when the resolver reads from the db, if the restriction is null then
 * it creates a new Restriction() object with null fields
 * In order to hide the meaningless restrictions in the API we filter them out here.
 * @see org.atlasapi.persistence.media.entity.VersionTranslator
 */
    private boolean containsMeaningfulData(Restriction restriction) {
```

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
pawel-mtbcst: we miss team-wide automated code formatting styling.

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
pawel-mtbcst: Why do we insist on copying collections at every possible place?

PR #717, opened 2022-03-14
(ENG-1217) Improve performance of CassandraEquivalentContentStore

tristanbeardmore-gray:
Refactors logic to only query for content that is likely to be needed by traversing the equivalence graph before fetching content

Adjusts logic of EquivalenceGraphFilter to remove unintuitive behaviour of selecting nodes that don't belong to the requested graph sources


tristanbeardmore-gray:
Abandoned in favour of ENG-1280

PR #730, opened 2022-06-28
ENG-1331

IvanKaraslavov:
Deprecating EBS


atlas-api/src/main/java/org/atlasapi/query/v4/schedule/EquivalentScheduleQueryExecutor.java
```diff
@@ -98,39 +98,11 @@ public QueryResult<ChannelSchedule> execute(ScheduleQuery query)
...
-                        .collect(MoreCollectors.toImmutableList());
-            } else {
-                channelSchedules = orderedChannelSchedules;
-            }
             return QueryResult.listResult(
-                    channelSchedules,
+                    orderedChannelSchedules,
                     query.getContext(),
-                    channelSchedules.size()
+                    ((List<ChannelSchedule>) orderedChannelSchedules).size()
```
tristanbeardmore-gray, line 105: This cast shouldn't be necessary

tristanbeardmore-gray (approved):

PR #735, opened 2022-10-06
Fixes case where lcn_sharing is not taken into account

EmiEnache:
Fixes channel_groups_summary to not ignore lcn_sharing

BT have a channel group with channels sharing lcn. But they're an
old customer and are using an annotation which we've not kept up
support of very well (channel_groups_summary). When the 
lcn_sharing annotation was introduced that old annotation wasn't 
updated to take it into account as a potential option.


EmiEnache:
They were trying to look at this:
https://voila.metabroadcast.com/4/channel_groups.json?ids=hnw9&annotations=channel_groups_summary,lcn_sharing&key=b4d2edb68da14dfb9e47b5465e99b1b1

tristanbeardmore-gray (approved):

PR #738, opened 2022-11-15
ENG-1344 SVOD Enrichment Atlas Model Changes

pawel-mtbcst:


atlas-cassandra/src/main/java/org/atlasapi/content/v2/CqlContentStore.java
```diff
@@ -101,6 +101,8 @@ protected CqlContentStore(Builder builder) {
         mappingManager.udtCodec(org.atlasapi.content.v2.model.udt.LocationSummary.class);
         mappingManager.udtCodec(org.atlasapi.content.v2.model.udt.LocationSummary.class);
         mappingManager.udtCodec(org.atlasapi.content.v2.model.udt.ItemSummary.class);
+        mappingManager.udtCodec(org.atlasapi.content.v2.model.udt.Company.class);
+        mappingManager.udtCodec(org.atlasapi.content.v2.model.udt.ProductionCompany.class);
```
tristanbeardmore-gray, line 105: Do we need entries for other types like Production? Don't remember exactly what this class needs.

tristanbeardmore-gray (changes requested):

atlas-cassandra/src/main/java/org/atlasapi/content/v2/model/udt/ProductionCompany.java
```diff
@@ -0,0 +1,36 @@
...
+    public void setCompany(Company company) {
+        this.company = company;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        ProductionCompany that = (ProductionCompany) o;
+        return company.equals(that.company);
```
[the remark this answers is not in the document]
pawel-mtbcst: This should never be null but as it's immutable and we have to provide default constructor, we may as well assume that empty collection is the same as null.

atlas-cassandra/src/main/java/org/atlasapi/content/v2/CqlContentStore.java
pawel-mtbcst: Yes, I wanted to show it to you guys on eng hh but forgot.
So this codec is able to transitively scan types but only 1 level in depth, meaning it will Scan `Production` and `Distribution` as it's referenced in `Content`, but it won't scan `ProductionCompany` and `Company` as it's too deep.

pawel-mtbcst: took a while to understand why it's not working tbh ...

PR #741, opened 2022-12-08
Merging encoding locations via providers

pawel-mtbcst:


atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -743,18 +746,37 @@ private <T extends Item> void mergeBroadcasts(
...
+                stream(orderedContent.spliterator(), false)
+                        .map(Content::getManifestedAs)
+                        .filter(Objects::nonNull)
+                        .flatMap(Set::stream)
+        )
+                .distinct()
+                .reduce(new HashSet<>(), this::accumulateEncoding, this::mergeEncodingAccumulators);
+
         chosen.setManifestedAs(encodings);
     }
```
EmiEnache, line 762: This is a little hard to make sense of and understand what it does. I'd improve the code readability, and also leave a comment explaining in natural language what the method does.

pawel-mtbcst: I'm really an enemy of putting comments on the method level describing what it does cause all it means is that code is not readable enough + they really tend to age rapidly, getting incoherent with what the code does as it's very easy not to maintain them causing a lot of confusion in future.

Would something like this make it more readable for you:
```
    private <T extends Content> void mergeEncodings(T chosen, Iterable<T> orderedContent) {
        Stream<Encoding> chosenEncodings = ofNullable(chosen.getManifestedAs())
                .map(Collection::stream)
                .orElse(Stream.empty());
        Stream<Encoding> orderedContentEncodings = stream(orderedContent.spliterator(), false)
                .map(Content::getManifestedAs)
                .filter(Objects::nonNull)
                .flatMap(Set::stream);

        Set<Encoding> encodings = concat(chosenEncodings, orderedContentEncodings)
                .distinct()
                .reduce(new HashSet<>(), this::accumulateEncoding, this::mergeEncodingAccumulators);

        chosen.setManifestedAs(encodings);
    }

```

atlas-core/src/main/java/org/atlasapi/content/Encoding.java
```diff
@@ -84,6 +85,12 @@ public boolean removeAvailableAt(Location location) {
         return this.availableAt.remove(location);
     }
 
+    public Set<Provider> locationProviders() {
```
EmiEnache, line 88: I find this name a little misleading; the method doesn't return the providers of a location, instead it returns the providers found in the locations contained within this encoding.
Perhaps rename to `getProviders`?

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
EmiEnache: Since it's hard to judge if a code is readable enough (we can understand code we've written, but can others? And who are the others - engineers in all likelihood, but what level of experience do they have? Can a graduate engineer read this and understand what it does? Lots of things to keep in mind) I find that in most cases where we don't expect the code to change much it helps to have a note or two on what the code does. That way any one can quickly read that instead of spending time understanding what the code does every time someone asks how it works. I agree that there's a risk of the comment becoming incoherent with the code, but I find that that's part of maintaining/refactoring code – if the comment is well located, then it will be easy to see it and update it if behaviour is updated. I will concede that lazy developers might not maintain them, although I don't like coding with that perspective in mind, and I think having a comment can save a lot of time and hassle in the future, and maintaining it doesn't take a lot of effort.

Having said all that, I think this is good enough now here, but the encoding accumulating is also so succinct that all in all it still results in this being a bit too arcane for my liking. 
To put it another way: it took me almost 30 minutes of reading this code to ensure it does what we intend it to do, and I'm only 90% sure it does. And I'm one of the people who participated in the conversation discussing what this change should look like! Let's say in 2024 we have an engineer who's never looked at this code before – how long do you think it's going to take them to answer a colleague's question of "how do VODs get merged in equived content in deer?" if no one else is available to assist with that answer?

pawel-mtbcst: @EmiEnache 

First of all - you don't know what code is prone to change and which is not. You can guess, but it's really hard to be accurate with that because there are several factors involved, some of them may occur in future - business decisions that will require reimplementing the logic, new requirements that you're not aware of yet, another feature's that may be indirectly affected by code you're working on currently, refactorings applied at some point and a lot of others. 
Again - if you need to add a note, it simply means that the code is not readable enough. The solution is not to write a comment - the solution is to make the code readable enough so that you don't need it. 

Second of all - it took you 30 minutes not because you wanted to understand what the code does but because you wanted to know if it does it right - the comment cannot say you that - only the tests do. Furthermore, chances are that some of this time you had to spend understanding the `reduce` pattern used on stream - that's an investment that will benefit you in the future. If I would use the same technique next time, you would read it faster I presume. 

Third - if you still feel a need to document the behavior for various conditions/inputs - simply write proper unit tests that serve as great documentation for developers that never gets out of date (cause when they start failing - you need to revisit either the test itself or the logic if the change was required).
I would agree that some comments may be beneficial if they tell you WHY the code has to look that way (more on that in the links attached), but not to repeat the code itself. I would also agree that some very specific and very complex blocks could be commented, but this is not the case. 

Fourth: if you need to quickly understand what the code does - simply read the method name. If the logic inside it is too long for the method name to capture it all - it's a clear sign that your method is too long and you need to refactor it - extract some lower-level finer-grained methods (which will also help you in testing BTW)
Fourth
> Most of it is never used because developers recognize that the documentation is probably not accurate, complete or up-to-date. And, even if the documentation did meet all of those criteria, there's no way for a developer to tell. It's safer to look at the reality of the executing code.

https://visualstudiomagazine.com/articles/2013/06/01/roc-rocks.aspx





And to answer your question:
> how do VODs get merged in equived content in deer?
He simply reads the code on a level that he's interested in. 

Btw. If a graduate engineer (or any other for that matter) cannot read the code, it's not necessarily the code's fault.

atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -743,18 +745,37 @@ private <T extends Item> void mergeBroadcasts(
...
-        HashSet<Encoding> encodings = Sets.newHashSet();
-        if (chosen.getManifestedAs() != null) {
-            encodings.addAll(chosen.getManifestedAs());
-        }
-        for (T item : orderedContent) {
-            if (item.getManifestedAs() != null) {
-                encodings.addAll(item.getManifestedAs());
-            }
-        }
+        Stream<Encoding> chosenEncodings = ofNullable(chosen.getManifestedAs())
```
tristanbeardmore-gray, line 748: See comments on slack, we can't just collapse all the encodings into a single stream since if a single piece of content has more than one location with the same provider we still need to retain all of them.

The code needs to be something more like this (untested):
```
        Set<Encoding> encodings = new HashSet<>();
        Set<Provider> seenProviders = new HashSet<>();
        Iterable<T> orderedContentWithChosen =
                Iterables.concat(ImmutableList.of(chosen), orderedContent);
        for (T content : orderedContentWithChosen) {
            Set<Provider> newProviders = new HashSet<>();
            if (content.getManifestedAs() != null) {
                for (Encoding encoding : content.getManifestedAs()) {
                    if (encoding.getLocationsProviders().stream()
                            .noneMatch(seenProviders::contains)) {
                        encodings.add(encoding);
                        newProviders.addAll(encoding.getLocationsProviders());
                    }
                }
            }
            seenProviders.addAll(newProviders);
        }
        chosen.setManifestedAs(encodings);
        ```

**Review by tristanbeardmore-gray — CHANGES_REQUESTED** [review]

File: atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java [hunk 4]
```diff
@@ -743,18 +746,37 @@ private <T extends Item> void mergeBroadcasts(
...
+
+        Set<Provider> alreadyAccumulatedProviders = acc.stream()
+                .flatMap(Encoding::getLocationsProviders)
+                .collect(toSet());
+
+        boolean anyProviderAlreadyAccumulated = encodings.stream()
+                .flatMap(Encoding::getLocationsProviders)
+                .anyMatch(alreadyAccumulatedProviders::contains);
+
+        if(anyProviderAlreadyAccumulated)
```
tristanbeardmore-gray, line 769: I still don't think this is correct, we still want to add some of these encodings if all the encoding's providers have not been seen before.

tristanbeardmore-gray (changes requested):

tristanbeardmore-gray (approved):

PR #743, opened 2023-01-27
Eng 1527

EmiEnache:
Non-film content didn't seem to have any merging logic for the `people` field. 
This change makes the merging logic not discriminate based on content type


atlas-core/src/main/java/org/atlasapi/output/OutputContentMerger.java
```diff
@@ -88,8 +91,8 @@ public class OutputContentMerger implements EquivalentsMergeStrategy<Content> {
 
     private static final Predicate<Item> HAS_BROADCASTS = input -> input.getBroadcasts()!= null && !input.getBroadcasts().isEmpty();
 
-    private static final Predicate<Film> HAS_PEOPLE =
-            film -> film.getPeople() != null && !film.getPeople().isEmpty();
+    private static final Predicate<Content> HAS_PEOPLE =
+            film -> film.people() != null && !film.people().isEmpty();
```
tristanbeardmore-gray, line 95:
```suggestion
            content -> content.people() != null && !content.people().isEmpty();
```

tristanbeardmore-gray (approved):

